Issue #1, opened 2019-02-13
SSR Support


bmullan91:
- `graphql-hooks-ssr` module
- Should support nested queries
- Should handle ssr errors
- Should pre-populate cache to be used for rehydration on the client

Issue #2, opened 2019-02-13
Replace graphql-request


bmullan91:
- Better error handling support
- Document how to pass in fetch / polyfill on node

bmullan91:
Fixed in #5

PR #5, opened 2019-02-14
Replaced graphql-request + refactored error handling

bmullan91:
Fixes: #1, #2 

### ⚠️ Breaking Changes ⚠️ 

- `createClient` -> `GraphQLClient` class
  - config changes
  - `endpoint` -> `url`
  - `fetch`: optional function to specify the fetch implementation | defaults to global fetch

### Additions
- New config options to `new GraphQLClient(config)`:
  - `logErrors`: Boolean | default __true__
  - `onError`: Function - custom error handler will receive ({ result, operation }) as the argument
- New properties available on the `result` object from all hooks:
  - `fetchError`: if an error occurred during the fetch call it will be available here
  - `httpError`: if a http error occurred it will be available here
  - `graphQLErrors`: The array of GraphQL errors from the server response
  - `error`: Boolean set to true if any of the above errors have been set
  - `cacheHit`: Boolean set to true if the result came from the cache, useful for debugging
- `operationName` can now be set:
  - `useQuery(query, { variables: {...}, operationName: 'MyOperation' })`
  - `useMutation(mutation, { variables: {...}, operationName: 'MyOperation' })`
  - `const [myMutation] = useMutation(mutation); myMutation({ variables: {}, operationName: 'MyOperation' })`


src/GraphQLClient.js
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  setHeaders(headers) {
+    this.headers = headers;
+    return this;
+  }
+
+  logErrorResult({ result, operation }) {
+    if (this.onError) {
+      return this.onError(result);
```
bmullan91: TODO: should be:

```js
return this.onError({ result, operation });
```

src/GraphQLClient.js
```diff
@@ -0,0 +1,131 @@
...
+        result = this.generateResult({
+          graphQLErrors: errors,
+          data
+        });
+      }
+    } catch (error) {
+      result = this.generateResult({
+        fetchError: error
+      });
+    } finally {
```
bmullan91: Does this _need_ to use `finally` 🤔

src/useQuery.js
```diff
@@ -15,9 +15,13 @@ module.exports = function useQuery(query, opts = {}) {
     ...opts
   });
 
+  // TODO revise this logic
```
bmullan91: Yea I think we can first check if `state.data || state.error` exists instead - if the query was already SSR'd then the initialState will be the cacheHit.

src/GraphQLClient.js
```diff
@@ -0,0 +1,131 @@
+class GraphQLClient {
+  constructor(config) {
+    this.cache = config.cache;
+    this.headers = config.headers || [];
```
jackdclark: Should this be an object? We're treating it as one in the rest of the file
```suggestion
    this.headers = config.headers || {};
```

src/useClientRequest.js
```diff
@@ -2,104 +2,101 @@ const React = require('react');
...
-
-  function getCacheHit(key) {
-    return useCache && !opts.skipCache && client.cache
-      ? client.cache.get(key) || null
-      : null;
-  }
+  const cacheKey = client.getCacheKey(operation, initialOpts);
+  const intialCacheHit = initialOpts.skipCache
+    ? null
+    : client.cache.get(cacheKey);
```
jackdclark: Do we need to make `cache` a required option to be passed in? Or should we provide a default cache, or default to no cache?

src/GraphQLClient.js
```diff
@@ -0,0 +1,131 @@
...
+    this.ssrMode = config.ssrMode;
+    this.ssrPromises = [];
+    this.url = config.url;
+    this.fetch = config.fetch;
```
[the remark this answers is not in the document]
bmullan91: Yea we should throw

src/useClientRequest.js
bmullan91: Cache is optional - this should check if client.cache exists.

README.md
```diff
@@ -26,10 +26,11 @@ or
...
 
 ```js
-import { createClient, ClientContext } from 'graphql-hooks';
+import { GraphQLClient, ClientContext } from 'graphql-hooks';
 
-const client = createClient({
-  endpoint: '/graphql'
+const client = new GraphQLClient({
+  url: '/graphql',
+  fetch // pass your preferred polyfill here
```
jackdclark: Maybe state that it's optional?

Issue #9, opened 2019-02-15
Add unit tests


jackdclark:
Add unit test coverage for all files, there are currently none

```
---------------------|----------|----------|----------|----------|-------------------|
File                 |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
---------------------|----------|----------|----------|----------|-------------------|
All files            |    17.43 |        0 |        0 |    17.59 |                   |
 ClientContext.js    |      100 |      100 |      100 |      100 |                   |
 GraphQLClient.js    |     1.85 |        0 |        0 |     1.89 |... 29,134,135,138 |
 index.js            |      100 |      100 |      100 |      100 |                   |
 useClientRequest.js |    14.29 |        0 |        0 |    14.29 |... 0,91,94,99,102 |
 useQuery.js         |    27.78 |        0 |        0 |    27.78 |... 27,28,29,33,35 |
---------------------|----------|----------|----------|----------|-------------------|```

**Comment by jackdclark** [comment]
We have good coverage for all files in the library now & [Coveralls](https://coveralls.io/github/nearform/graphql-hooks?branch=master) reporting in CI to make sure this doesn't decrease. Closing this.

```
---------------------|----------|----------|----------|----------|-------------------|
File                 |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
---------------------|----------|----------|----------|----------|-------------------|
All files            |      100 |    98.63 |      100 |      100 |                   |
 ClientContext.js    |      100 |      100 |      100 |      100 |                   |
 GraphQLClient.js    |      100 |    97.56 |      100 |      100 |                85 |
 index.js            |      100 |      100 |      100 |      100 |                   |
 useClientRequest.js |      100 |      100 |      100 |      100 |                   |
 useQuery.js         |      100 |      100 |      100 |      100 |                   |
---------------------|----------|----------|----------|----------|-------------------|
```

PR #24, opened 2019-02-19
Unit tests for useQuery

jackdclark:
Part of #9 

Add test coverage for `useQuery`

```
 PASS  test/unit/useQuery.test.js
  useQuery
    ✓ calls useClientRequest with query (22ms)
    ✓ calls useClientRequest with options (2ms)
    ✓ returns initial state from useClientRequest & refetch (2ms)
    ✓ bypasses cache when refetch is called (1ms)
    ✓ sends the query on mount if no data & no error (2ms)
    ✓ does not send the same query on mount if data is already present (2ms)
    ✓ does not send the query on mount if there is an error (1ms)
    ✓ adds query to ssrPromises when in ssrMode if no data & no error (3ms)
    ✓ does not add query to ssrPromises when in ssrMode if there is already data (1ms)
    ✓ does not add query to ssrPromises when in ssrMode if there is an error (1ms)
    ✓ does not add query to ssrPromises when in ssrMode if ssr is overridden in options (2ms)
    ✓ does not send the same query twice (2ms)
    ✓ sends the query again if the variables change (5ms)
    ✓ sends another query if the query changes (2ms)
```

### DIFF
--- a/test/unit/useQuery.test.js
+++ b/test/unit/useQuery.test.js
@@ -1,5 +1,104 @@
-import { useQuery } from '../../src';
+import React from 'react';
+import { testHook } from 'react-testing-library';
+import { ClientContext, useQuery } from '../../src';
+
+let mockQueryReq, mockState;
+jest.mock('../../src/useClientRequest', () => () => [mockQueryReq, mockState]);
+
+let mockClient;
+const Wrapper = props => (
+  <ClientContext.Provider value={mockClient}>
+    {props.children}
+  </ClientContext.Provider>
+);
+
+const TEST_QUERY = `query Test($limit: Int) {
+  tests(limit: $limit) {
+    id
+  }
+}`;
 
 describe('useQuery', () => {
-  it('runs a test', () => {});
+  beforeEach(() => {
+    mockQueryReq = jest.fn();
+    mockState = { loading: true, cacheHit: false };
+    mockClient = {
+      ssrMode: false,
+      ssrPromises: []
+    };
+  });
+
+  afterEach(() => {
+    jest.unmock('../../src/useClientRequest');
+  });
+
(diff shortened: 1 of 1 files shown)


test/unit/useQuery.test.js
```diff
@@ -1,5 +1,123 @@
...
+    expect(mockQueryReq).toHaveBeenCalledTimes(1);
+  });
+
+  it('does not send the same query on mount if data is already present', () => {
+    mockState.data = { some: 'data' };
+    testHook(() => useQuery(TEST_QUERY), { wrapper: Wrapper });
+    expect(mockQueryReq).not.toHaveBeenCalled();
+  });
+
+  it('does not send the query on mount if there is already data', () => {
```
bmullan91:
```suggestion
  it('does not send the query on mount if there is already an error', () => {
```

test/unit/useQuery.test.js
```diff
@@ -1,5 +1,123 @@
...
+    mockState.error = true;
+    testHook(() => useQuery(TEST_QUERY), { wrapper: Wrapper });
+    expect(mockQueryReq).not.toHaveBeenCalled();
+  });
+
+  it('does not send the query on mount if there is an error', () => {
```
bmullan91: Ah this is the correct test - the one above is a duplicate.

jackdclark: Nice spot, removed this

jackdclark:
@Joezo @bmullan91 thanks for the reviews. I've added the following three tests to cover the `useEffect` conditional firing
```
    ✓ does not send the same query twice (2ms)
    ✓ sends the query again if the variables change (2ms)
    ✓ sends another query if the query changes (2ms)
```

PR #30, opened 2019-02-19
unit tests for useClientRequest

jackdclark:
Part of #9 

Adds unit tests for `useClientRequest`

Note: can't currently test `fetchData`, [see explanation here](https://github.com/nearform/graphql-hooks/blob/9a09e6ca154a03f44855c4684d51ad96988ecf4b/test/unit/useClientRequest.test.js#L76-L94). Raised an issue of this #31.

```
 PASS  test/unit/useClientRequest.test.js
  useClientRequest
    ✓ returns a fetch function & state (21ms)
    initial state
      ✓ includes the cached response if present (2ms)
      ✓ skips the cache if skipCache is passed in (2ms)
      ✓ skips the cache if a cache is not configured (2ms)
    fetchData
      ○ skipped 1 test
```

### DIFF
--- a/test/unit/useClientRequest.test.js
+++ b/test/unit/useClientRequest.test.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import { act, testHook } from 'react-testing-library';
 import { useClientRequest, ClientContext } from '../../src';
-import { pathToFileURL } from 'url';
 
 let mockClient;
 
--- a/test/unit/useClientRequest.test.js
+++ b/test/unit/useClientRequest.test.js
@@ -40,8 +40,8 @@ describe('useClientRequest', () => {
   describe('initial state', () => {
     it('includes the cached response if present', () => {
       mockClient.cache.get.mockReturnValueOnce({ some: 'cached data' });
-      let _, state;
-      testHook(() => ([_, state] = useClientRequest(TEST_QUERY)), {
+      let state;
+      testHook(() => ([, state] = useClientRequest(TEST_QUERY)), {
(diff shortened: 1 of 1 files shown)


test/unit/useClientRequest.test.js
```diff
@@ -1,5 +1,97 @@
-import { useClientRequest } from '../../src';
+import React from 'react';
+import { act, testHook } from 'react-testing-library';
+import { useClientRequest, ClientContext } from '../../src';
+import { pathToFileURL } from 'url';
```
Joezo: Is this meant to be here? Can't see it used.

test/unit/useClientRequest.test.js
```diff
@@ -1,5 +1,97 @@
...
+    });
+    expect(fetchData).toEqual(expect.any(Function));
+    expect(state).toEqual({ cacheHit: false, loading: true });
+  });
+
+  describe('initial state', () => {
+    it('includes the cached response if present', () => {
+      mockClient.cache.get.mockReturnValueOnce({ some: 'cached data' });
+      let _, state;
+      testHook(() => ([_, state] = useClientRequest(TEST_QUERY)), {
```
Joezo: You don't need the `_` assignment here. You can do `[, state] = useClient...`.

Same thing on further down tests.

test/unit/useClientRequest.test.js
jackdclark: Nope, my only explanation is VSCode must have added it without me knowing 😬. Thanks!

PR #32, opened 2019-02-20
Add useClientRequest fetchData tests with warnings

### LINKED ISSUE #31: Test fetchData from useClientRequest
See https://github.com/nearform/graphql-hooks/blob/9a09e6ca154a03f44855c4684d51ad96988ecf4b/test/unit/useClientRequest.test.js#L76-L94

jackdclark:
Implements the tests that were left out of #30 
Resolves #31 

```
    fetchData
      ✓ calls request with options & updates the state with the result (10ms)
      ✓ calls request with revised options (2ms)
      ✓ skips the request & returns the cached data if it exists (2ms)
      ✓ skips the cache if skipCache is passed in (2ms)
      ✓ skips the cache if skipCache is there is no cache (2ms)
      ✓ sets the result from the request in the cache (2ms)
```

The following warning will display in the console for each test until https://github.com/facebook/react/issues/14769 has been resolved. This is because `fetchData` updates the state & should be wrapped inside an `act`, however `act` does not yet support `async/await` & `fetchData` is an async function.

```
  console.error node_modules/react-dom/cjs/react-dom.development.js:506
    Warning: An update to TestHook inside a test was not wrapped in act(...).

    When testing, code that causes React state updates should be wrapped into act(...):

    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */

    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
        in TestHook
        in Wrapper
```

### DIFF
--- a/test/unit/useClientRequest.test.js
+++ b/test/unit/useClientRequest.test.js
@@ -25,7 +25,7 @@ describe('useClientRequest', () => {
         get: jest.fn(),
         set: jest.fn()
       },
-      request: jest.fn().mockResolvedValue('bla')
+      request: jest.fn().mockResolvedValue({ some: 'data' })
     };
   });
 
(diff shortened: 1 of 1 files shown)


test/unit/useClientRequest.test.js
```diff
@@ -73,25 +73,112 @@ describe('useClientRequest', () => {
   });
 
   describe('fetchData', () => {
-    it.skip('needs testing', () => {
-      // fetchData updates the state & therefore needs to be wrapped inside an `act`
```
Joezo: I would probably leave this comment in. Someone is inevitably going to look at this test when they see it kicking out warnings so it'd be good to leave this info here.

jackdclark: Good shout

PR #33, opened 2019-02-20
Unit tests for useManualQuery

jackdclark:
Part of #9 

Adds a test for `useManualQuery` which is just a proxy to `useClientRequest` with `useCache=true`

```
 PASS  test/unit/useManualQuery.test.js
  useManualQuery
    ✓ calls useClientRequest with useCache set to true & options (5ms)
```

### DIFF
--- a/test/unit/useManualQuery.test.js
+++ b/test/unit/useManualQuery.test.js
@@ -9,10 +9,6 @@ const TEST_QUERY = `query Test($limit: Int) {
 }`;
 
 describe('useManualQuery', () => {
-  afterEach(() => {
-    jest.unmock('../../src/useClientRequest');
-  });
-
   it('calls useClientRequest with useCache set to true & options', () => {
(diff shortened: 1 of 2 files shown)


test/unit/useManualQuery.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+const TEST_QUERY = `query Test($limit: Int) {
+  tests(limit: $limit) {
+    id
+  }
+}`;
+
+describe('useManualQuery', () => {
+  afterEach(() => {
+    jest.unmock('../../src/useClientRequest');
```
Joezo: Should the `mock` be in a `beforeEach`? I'm unsure on the inner workings of jest but it reads like after the first test is run in this file then the imported file is unmocked for the remaining ones.

jackdclark: I think this was a misunderstanding from myself on how the mocks work, removing it has no impact. Will remove.

PR #39, opened 2019-02-21
Fix initial loading state with useMutation and add optmization

Joezo:
This makes sure that `loading` is set to false when `useMutation` is called. It causes issues having `loading` default to true for a mutation because it doesn't actually send the request until the consumer calls the returned method. 

This is probably considered a breaking change? @bmullan91 / @jackdclark 

Another small thing in this PR is an additional check to avoid making a new state object if we don't need to, this means react won't try to render wastefully.

### DIFF
--- a/src/useClientRequest.js
+++ b/src/useClientRequest.js
@@ -53,7 +53,7 @@ function useClientRequest(query, initialOpts = {}) {
   const [state, dispatch] = React.useReducer(reducer, {
     ...intialCacheHit,
     cacheHit: !!intialCacheHit,
-    loading: !intialCacheHit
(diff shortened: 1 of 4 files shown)


src/useClientRequest.js
```diff
@@ -48,12 +51,12 @@ function useClientRequest(query, initialOpts = {}) {
   };
 
   const cacheKey = client.getCacheKey(operation, initialOpts);
-  const intialCacheHit =
+  const initialCacheHit =
```
bmullan91: Ah, my poor spelling strikes again 🎯

PR #44, opened 2019-02-21
Pagination

### LINKED ISSUE #29: Add pagination helper
Whilst implementing an example app in next.js it occurred that pagination is pretty complicated and could be simplified by `useQuery` offering a method to help load more data and update the existing data without having to refetch everything each time.
Apollo does this with a `fetchMore` method https://www.apollographql.com/docs/react/features/pagination.html#fetch-more

jackdclark:
Closes: #41, closes #42, and closes #29 

~Provides a `fetchMore` function that will send the same query with the provided options merged into the original options, meaning the consumer doesn't have to manually maintain the state.~ Removed this as the same can be achieved with `refetch` now that it accepts `options`.

`useQuery` accepts an `updateData` function as part of the `options` object that enables control of how the previous & new results are merged, this function can be overridden if necessary when calling `refetch` as well.

### TODO
- [x] Agree on implementation
- [x] Test coverage
- [x] Update documentation

### Example usage

```jsx
// use options.updateResult to append the new page of posts to our current list of posts
const updateData = (prevData, prevData) => ({
  ...data,
  allPosts: [...prevData.allPosts, ...data.allPosts]
});

export default function PostList() {
  // set a default offset of 0 to load the first page
  const [skipCount, setSkipCount] = useState(0);

  const { loading, error, data } = useQuery(
    allPostsQuery,
    { variables: { skip: skipCount, first: 10 } },
    updateData
  );

  if (error) return <div>There was an error!</div>;
  if (loading && !data) return <div>Loading</div>;

  const { allPosts, _allPostsMeta } = data;
  const areMorePosts = allPosts.length < _allPostsMeta.count;

  return (
    <section>
      <ul>
        {allPosts.map(post => (
          <li key={post.id}>
            <a href={post.url}>{post.title}</a>
          </li>
        ))}
      </ul>
      {areMorePosts && (
        <button
          // set the offset to the current number of posts to fetch the next page
          onClick={() => setSkipCount(allPosts.length)}
        >
          Show more
        </button>
      )}
    </section>
  );
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,8 +201,8 @@ const { loading, error, data, refetch, cacheHit, ...errors } = useQuery(QUERY);
 - `loading`: Boolean - `true` if the query is in flight
 - `error`: Boolean - `true` if `fetchError` or `httpError` or `graphQLErrors` has been set
 - `data`: Object - the result of your GraphQL query
-- `refetch(options)`: Function - useful when refetching the same query after a mutation; NOTE this presets `skipCache=true`
-  - options: Object - options that will be merged into the `options` that were passed into `useQuery` (see above).
+- `refetch(options)`: Function - useful when refetching the same query after a mutation; NOTE this presets `skipCache=true` & will bypass the `options.updateData` function that was passed into `useQuery`. You can pass a new `updateData` into `refetch` if necessary.
+  - `options`: Object - options that will be merged into the `options` that were passed into `useQuery` (see above).
 - `cacheHit`: Boolean - `true` if the query result came from the cache, useful for debugging
 - `fetchError`: Object - Set if an error occured during the `fetch` call
 - `httpError`: Object - Set if an error response was returned from the server
--- a/README.md
+++ b/README.md
@@ -187,6 +187,9 @@ This is a custom hook that takes care of fetching your query and storing the res
   - `skipCache`: Boolean - defaults to `false`; If `true` it will by-pass the cache and fetch, but the result will then be cached for subsequent calls. Note the `refetch` function will do this automatically
   - `ssr`: Boolean - defaults to `true`. Set to `false` if you wish to skip this query during SSR
   - `fetchOptionsOverrides`: Object - Specific overrides for this query. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch) for info on what options can be passed
+  - `updateResult(previousResult, result)`: Function - Custom handler for merging previous & new query results; return value will replace `data` in `useQuery` return value
+    - `previousResult`: Previous GraphQL query or `updateResult` result
+    - `result`: New GraphQL query result
 
 ### `useQuery` return value
 
@@ -198,6 +201,8 @@ const { loading, error, data, refetch, cacheHit, ...errors } = useQuery(QUERY);
 - `error`: Boolean - `true` if `fetchError` or `httpError` or `graphQLErrors` has been set
 - `data`: Object - the result of your GraphQL query
 - `refetch`: Function - useful when refetching the same query after a mutation; NOTE this presets `skipCache=true`
+- `fetchMore(options)`: Function - send the same query with updated options, useful for pagination.
+  - `options`: Object - options that will be merged into the `options` that were passed into `useQuery` (see above).
 - `cacheHit`: Boolean - `true` if the query result came from the cache, useful for debugging
 - `fetchError`: Object - Set if an error occured during the `fetch` call
 - `httpError`: Object - Set if an error response was returned from the server
--- a/README.md
+++ b/README.md
@@ -188,9 +188,9 @@ This is a custom hook that takes care of fetching your query and storing the res
   - `skipCache`: Boolean - defaults to `false`; If `true` it will by-pass the cache and fetch, but the result will then be cached for subsequent calls. Note the `refetch` function will do this automatically
   - `ssr`: Boolean - defaults to `true`. Set to `false` if you wish to skip this query during SSR
   - `fetchOptionsOverrides`: Object - Specific overrides for this query. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch) for info on what options can be passed
-  - `updateResult(previousResult, result)`: Function - Custom handler for merging previous & new query results; return value will replace `data` in `useQuery` return value
-    - `previousResult`: Previous GraphQL query or `updateResult` result
-    - `result`: New GraphQL query result
+  - `updateData(previousData, data)`: Function - Custom handler for merging previous & new query results; return value will replace `data` in `useQuery` return value
+    - `previousData`: Previous GraphQL query or `updateData` result
+    - `data`: New GraphQL query result
 
 ### `useQuery` return value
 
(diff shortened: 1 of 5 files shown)


src/useClientRequest.js
```diff
@@ -57,7 +57,7 @@ function useClientRequest(query, initialOpts = {}) {
   });
 
   // arguments to fetchData override the useClientRequest arguments
-  async function fetchData(newOpts) {
+  async function fetchData(newOpts, updateResult) {
```
bmullan91: What if we made `updateResult` part of the `options` object?

src/useQuery.js
```diff
@@ -25,11 +23,21 @@ module.exports = function useQuery(query, opts = {}) {
   }
 
   React.useEffect(() => {
-    queryReq();
+    queryReq(null, updateResult);
```
bmullan91: If the `update` function was part of the `options` we wouldn't need to do this

src/useQuery.js
```diff
@@ -25,11 +23,21 @@ module.exports = function useQuery(query, opts = {}) {
...
+    fetchMore: (fetchMoreOpts, updateResultOverride) => {
+      queryReq(
+        {
+          ...allOpts,
+          ...fetchMoreOpts,
+          variables: { ...opts.variables, ...fetchMoreOpts.variables }
+        },
+        updateResultOverride || updateResult
+      );
+    }
```
bmullan91: This would become:

```js
fetchMore: (fetchMoreOptions) => queryReq(fetchMoreOptions)
```

We don't need to merge the options here - they are already bound to the `queryReq` by the closure in useClientRequest which does the merging/overriding.

src/useQuery.js
```diff
@@ -7,13 +7,11 @@ const defaultOpts = {
   useCache: true
 };
 
-module.exports = function useQuery(query, opts = {}) {
+module.exports = function useQuery(query, opts = {}, updateResult) {
```
bmullan91: If `update` was part of the options we don't need to update the `useQuery` signature.

src/useClientRequest.js
jackdclark: @Joezo suggested the same, will move it 👍

src/useQuery.js
jackdclark: In this merge I was also doing a deeper merge on `options.variables` so that only the ones that are changing need to be passed in each time. As discussed, I've updated this to only do the merge on the variables & pass the rest through.

src/useQuery.js
```diff
@@ -30,6 +28,10 @@ module.exports = function useQuery(query, opts = {}) {
 
   return {
     ...state,
-    refetch: () => queryReq({ skipCache: true })
+    refetch: () => queryReq({ skipCache: true }),
+    fetchMore: ({ variables, ...rest } = {}) =>
+      // merge variables so they don't all have to be passed back in,
+      // just the ones that are changing
+      queryReq({ ...rest, variables: { ...allOpts.variables, ...variables } })
```
bmullan91: 😍

test/unit/useClientRequest.test.js
```diff
@@ -212,7 +212,91 @@ describe('useClientRequest', () => {
...
+            })),
+          { wrapper: Wrapper }
+        );
+
+        // first fetch to populate state
+        await fetchData();
+
+        expect(fetchData({ variables: { limit: 20 } })).rejects.toThrow(
+          'options.updateResult must be a function'
+        );
```
bmullan91: 👏 excellent tests

jackdclark:
@bmullan91 as discussed earlier I've removed `fetchMore` in favour of the declarative implementation. If a more imperative style is desired then the same can be achieved by using `refetch` and passing in new options.

I've also added a pagination guide to the README that includes distinct page & infinite loading examples.

If you could take another look when you get a chance please, hopefully we can get this in next week 😀

src/useClientRequest.js
```diff
@@ -88,7 +88,14 @@ function useClientRequest(query, initialOpts = {}) {
...
 
     dispatch({ type: actionTypes.LOADING });
-    const result = await client.request(revisedOperation, revisedOpts);
+    let result = await client.request(revisedOperation, revisedOpts);
+
+    if (state.data && result.data && revisedOpts.updateResult) {
+      if (typeof revisedOpts.updateResult !== 'function') {
+        throw new Error('options.updateResult must be a function');
+      }
+      result.data = revisedOpts.updateResult(state.data, result.data);
```
bmullan91: Since the name is `updateResult`, maybe we should pass the entire result? It might be useful for a developer to change / override the errors in some circumstances.

`useClientRequest` would then look something like:
```js
let result = await client.request(revisedOperation, revisedOpts);

if (revisedOpts.updateResult) {
  if (typeof revisedOpts.updateResult !== 'function') {
    throw new Error('options.updateResult must be a function');
  }
  result = revisedOpts.updateResult(state, result)
}
```

The pagination example `updateResult`:
```js
const updateResult = (prevResult, newResult) => ({
  ...newResult,
  data: {
    ...newResult.data,
    allPosts: [...prevResult.data.allPosts, ...result.data.allPosts]
  }
});

useQuery(MY_QUERY, {
  updateResult
})
```

It makes updating the data a bit more verbose, but the developer could write a wrapper if it was a really common usecase:

```js
const dataUpdater = (fn) => (prevResult, newResult) => ({
  ...newResult,
  data: fn(prevResult.data, newResult.data)
})
```

jackdclark: As discussed, passing the entire result back allows internal properties like `loading` & `error` to be changed, or new ones to be added which could be abused or cause tricky bugs. We're going to rename this to `updateData` as a better description of what it's doing & if we see a use case for changing / overriding errors in the future, we could always introduce an `updateErrors` option.

README.md
```diff
@@ -309,6 +314,130 @@ The `options` object that can be passed either to `useMutation(mutation, options
...
+      title
+      votes
+      url
+      createdAt
+    }
+    _allPostsMeta {
+      count
+    }
+  }
+`;
```
bmullan91: We should simplify the example - this is could be:

```js
ALL_POSTS_QUERY = `
  query($limit: Int, $skip: Int) {
    allPosts(limit: $limit, skip: $skip) {
      total,
      items {
        id
        title
      }
    }
  }
`
```

src/useQuery.js
```diff
@@ -30,6 +28,11 @@ module.exports = function useQuery(query, opts = {}) {
 
   return {
     ...state,
-    refetch: () => queryReq({ skipCache: true })
+    refetch: (options = {}) =>
+      queryReq({
+        skipCache: true,
+        updateData: (_, data) => data,
```
bmullan91: Minor: We should add a comment here saying why we're doing this

README.md
jackdclark: I think the separation between data & meta is good here? There are redundant parts to this example that could be removed & make it more readable though, how about:

```javascript
export const ALL_POSTS_QUERY = `
  query allPosts($first: Int!, $skip: Int!) {
    allPosts(first: $first, skip: $skip) {
      id
      title
      url
    }
    _allPostsMeta {
      count
    }
  }
`;
```

src/useQuery.js
jackdclark: Agree, I've added an explanation

PR #63, opened 2019-02-27
Use rollup to generate bundles for UMD, ESM and CJS

### LINKED ISSUE #11: Create separate bundles
- umd
- cjs
- es
- mjs?

Joezo:
Co-authored-by: Jack Clark

### What does this PR do?

Generates bundles for UMD, ESM and CommonJS. This brings with it support for browsers >= IE11. 

Whilst working on this we published a few version to `npm` with the `test` tag so that it could be tested. This [codepen](https://codepen.io/anon/pen/XGbdxR) has been created which uses the `UMD` build. We spun up a quick Create react app to test out `ESM` which worked great. The nextjs example app has also been updated locally to use the commonjs version. 

Our rollup config was heavily influenced by the one [in redux](https://github.com/reduxjs/redux/blob/master/rollup.config.js).

Bundle size has slightly increased but we now have increased browser support and our module is now tree shakeable, meaning potentially reduced size for end users.

*Important*
We now use Promises to write this library as transpiling from async/await is usually avoided in libraries as it's expensive. We opted instead to use untranspiled Promises and then inform the user they should provide an environment that includes Promises.

### Related issues

Resolves #11 

### Checklist


package.json
```diff
@@ -28,7 +37,7 @@
   },
   "devDependencies": {
     "@babel/preset-env": "7.3.4",
-    "@babel/preset-react": "7.0.0",
+    "@babel/preset-react": "^7.0.0",
```
bmullan91: @Joezo Could you pin these devDeps - @renovate[bot] will shout at you 🙈

src/.babelrc.js
```diff
@@ -0,0 +1,22 @@
...
+
+module.exports = {
+  presets: [
+    [
+      '@babel/env',
+      {
+        targets: {
+          browsers: ['ie >= 11']
+        },
+        exclude: ['transform-async-to-generator', 'transform-regenerator'],
```
bmullan91: I noticed that `async/await` wasn't transformed - I guess this is whats doing that. Does this mean that consumers need to transpile this when they bundle their app?

bmullan91: I've added the following https://github.com/nearform/graphql-hooks/issues/11#issuecomment-467922037 RE the decision to replace `async/await` with Promises.

README.md
```diff
@@ -25,6 +25,16 @@ or
 
 `yarn add graphql-hooks`
 
+## Support
+
+- Node.js v8.5+
```
bmullan91: I don't think this is accurate, this _should_ run on v4

README.md
```diff
@@ -25,6 +25,16 @@ or
...
+- Browsers [`> 1%, not dead`](https://browserl.ist/?q=%3E+1%25%2C+not+dead)
+
+Consider polyfilling:
+
+- [`Promise`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Promise)
+- [`fetch`](https://developer.mozilla.org/docs/Web/API/Fetch_API)
```
bmullan91: We should link to the `GraphQLClient` docs where the user can pass in the `fetch` implementation instead of polyfilling.

package.json
```diff
@@ -2,15 +2,24 @@
...
   "scripts": {
     "test": "jest",
     "test:watch": "jest --watch",
     "test:coverage": "jest --coverage",
     "prettier": "pretty-quick",
     "contributors:add": "all-contributors add",
-    "contributors:generate": "all-contributors generate"
+    "contributors:generate": "all-contributors generate",
+    "build": "rollup -c",
+    "prepublishOnly": "npm run build"
```
bmullan91: We need to run `npm run build` in `prepack` also.

jackdclark: If we add it in `prepack` I believe it will run twice when we run `npm publish`. Are we using `npm pack`?

README.md
jackdclark: I'll do some local testing

package.json
Joezo: I think `npm pack` might be an edge case and if people want to use it they could do `npm run build` first?

README.md
jackdclark: As discussed we probably don't need this as we already detail what they need to polyfill, removing.

mcollina: Let's keep this. Nobody should not run on v4 anyway.

jackdclark: I've updated this to "Latest 8 & 10 Node releases"

Issue #64, opened 2019-02-27
Add eslint-plugin-react-hooks


bmullan91:
### Description

Add https://www.npmjs.com/package/eslint-plugin-react-hooks to ensure we're doing the right thing 😅.

Joezo:
This was completed as part of #77

PR #77, opened 2019-02-28
feat: integrate eslint-plugin-react-hooks

jh3y:
### What does this PR do?
- adds `eslint`
- switches to the recommended `lint-staged` to integrate with `prettier`
- pulls in `eslint-plugin-react-hook` 👍 
- formats files that needed formatting
- disables `no-console` for logging in the `GraphQLClient` file

### Related issues
#64 

### Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,9 @@ module.exports = {
   parser: 'babel-eslint',
   env: {
     browser: true,
-    es6: true
+    node: true,
+    es6: true,
+    jest: true
   },
   extends: [
     'eslint:recommended',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,37 @@
+module.exports = {
+  parser: 'babel-eslint',
+  env: {
+    browser: true,
+    es6: true
+  },
+  extends: [
+    'eslint:recommended',
+    'plugin:react/recommended',
+    'prettier',
+    'plugin:prettier/recommended'
+  ],
+  globals: {
+    Atomics: 'readonly',
+    SharedArrayBuffer: 'readonly'
+  },
+  parserOptions: {
+    ecmaFeatures: {
+      jsx: true
+    },
+    ecmaVersion: 2018,
+    sourceType: 'module'
+  },
+  settings: {
+    react: {
+      version: '16.8.3'
+    }
+  },
+  plugins: ['prettier', 'react', 'react-hooks'],
+  rules: {
+    'prettier/prettier': 'error',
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -9,7 +9,8 @@
     "test": "jest",
     "test:watch": "jest --watch",
     "test:coverage": "jest --coverage",
-    "prettier": "pretty-quick",
+    "eslint": "eslint src",
+    "prettier": "prettier src/**/*.js --write",
```
bmullan91: Same goes for tests - we should make sure they're pretty 💅

.eslintrc.js
```diff
@@ -0,0 +1,37 @@
...
+  settings: {
+    react: {
+      version: '16.8.3'
+    }
+  },
+  plugins: ['prettier', 'react', 'react-hooks'],
+  rules: {
+    'prettier/prettier': 'error',
+    'react-hooks/rules-of-hooks': 'error',
+    'react-hooks/exhaustive-deps': 'warn',
```
bmullan91: This should also be `error` - but if we're going to fix the issues in a separate PR we can set it to `error` then.

src/GraphQLClient.js
```diff
@@ -35,13 +35,13 @@ class GraphQLClient {
...
   }
-
+  /* eslint-disable no-console */
   logErrorResult({ result, operation }) {
     if (this.onError) {
       return this.onError({ result, operation });
     }
 
-    console.error(`GraphQL Hooks Error`);
+    console.error('GraphQL Hooks Error');
```
bmullan91: 👍

package.json
```diff
@@ -31,17 +32,23 @@
   "author": "Brian Mullan <bmullan91@gmail.com>",
   "license": "Apache-2.0",
   "peerDependencies": {
-    "react": "^16.8.1"
+    "react": "16.8.3"
```
bmullan91: `peerDependencies` should be be kept loose - please keep the `^`

jh3y:
Thanks for taking a look @bmullan91 😄 - No problem will make those changes tomorrow 👍

.eslintrc.js
jh3y: @bmullan91 - I haven't set this to `error` in this PR because it wanted me to change the deps in those two files. Changing as advised by the linter broke the tests or removed the deps array completely. Might be best for someone familiar with that behaviour to take a look at those warnings.

jh3y:
Updated. I've left a comment about the `exhaustive deps` 👍

PR #96, opened 2019-03-05
Handle unmounting whilst request is in flight

### LINKED ISSUE #54: Requests are not cancelled on unmount
### Environment

- `graphql-hooks` version: 3.1.0
- `react` version: 16.8.3
- Browser: Any

### Description

Requests are not cancelled on unmount which means we try to update state after component is gone.

screenshot 2019-02-25 at 14 51 44

### How to reproduce

Can be reproduced in the example next app by simulating slow connections and upvoting a post, then navigating to the about page before the request returns.

Keep track of whether the component is mounted by making useClientRequest return a method to cancel everything. This would switch an internal var that we would check once the request has returned, so that we don't then try to set state.

Joezo:
### What does this PR do?
Prevents `fetchData` within `useClientRequest` from updating state when a request returns, if the component has unmounted.

### Related issues

closes #54 

### Checklist

### DIFF
--- a/packages/graphql-hooks/test/unit/useClientRequest.test.js
+++ b/packages/graphql-hooks/test/unit/useClientRequest.test.js
@@ -205,6 +205,30 @@ describe('useClientRequest', () => {
       expect(state).toEqual({ cacheHit: false, loading: false, data: 'data' })
     })
 
+    it('calls request with options & doesnt update state if component unmounts whilst request is in flight', async () => {
+      let fetchData, state;
+      const { unmount } = renderHook(
+        () =>
+          ([fetchData, state] = useClientRequest(TEST_QUERY, {
+            variables: { limit: 2 },
+            operationName: 'test'
+          })),
+        {
+          wrapper: Wrapper
+        }
+      );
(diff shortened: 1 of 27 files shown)


packages/graphql-hooks/test/unit/useClientRequest.test.js
```diff
@@ -205,6 +205,30 @@ describe('useClientRequest', () => {
...
+            operationName: 'test'
+          })),
+        {
+          wrapper: Wrapper
+        }
+      )
+
+      const fetchDataPromise = fetchData()
+      unmount()
+      await fetchDataPromise
```
bmullan91: In this test it should resolve the result, right? We can then add a test that after unmount it resolves undefined:

```js
const fetchDataPromise = fetchData()
unmount()
const firstResult = await fetchDataPromise // <- expected actual result
const secondResult = await fetchData() // <- expected undefined
```

Joezo: I've added a test for that 👍

PR #104, opened 2019-03-06
Conventional commits

### LINKED ISSUE #48: Consider using semantic-release
## Notes
- commits will need to follow a certain format `fix`, `feat` or `BREAKING CHANGE`
- semantic-release figures out what the new version should be
- generates release notes
- generates git tag
- publish the new release

see https://github.com/semantic-release/semantic-release for more info.

jackdclark:
### What does this PR do?

- Adds [conventional commits Lerna configuration](https://github.com/lerna/lerna/blob/030de9d6285f2b364ed1460b9534bc2190acdf7a/commands/version/README.md#--conventional-commits)
- Adds [GitHub release Lerna configuration](https://github.com/lerna/lerna/blob/030de9d6285f2b364ed1460b9534bc2190acdf7a/commands/version/README.md#--github-release)
- Updates CONTRIBUTING.md 

### Related issues

Resolves #48 

### Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,8 +59,10 @@ For more information on `@all-contributors` see it's [usage docs](https://allcon
 
 ### Release process:
 
+When merging a pull request to `master`, the squashed commit message should follow the [Conventional Commits](https://www.conventionalcommits.org) specification. This enables us to automatically generate CHANGELOGs & determine a semantic version bump.
+
 - `npm test`
-- `NPM_CONFIG_OTP=<your-otp> npm run release`
+- `NPM_CONFIG_OTP=<your-otp> GH_TOKEN=<your-github-token> npm run release`
 
 Follow the prompts from [`lerna publish`](https://lernajs.io/#command-publish)
 
--- a/lerna.json
+++ b/lerna.json
@@ -4,6 +4,14 @@
   "command": {
     "bootstrap": {
       "npmClientArgs": ["--no-package-lock"]
+    },
+    "publish": {
+      "conventionalCommits": true,
+      "message": "chore(release): :rocket:"
+    },
+    "version": {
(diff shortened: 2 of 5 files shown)


Joezo:
Might be worth putting a link in the contributing docs which goes to the different commit type prefixes that are available.

jackdclark:
I've added a note to the CONTRIBUTING.md @Joezo

CONTRIBUTING.md
```diff
@@ -59,8 +59,10 @@ For more information on `@all-contributors` see it's [usage docs](https://allcon
 
 ### Release process:
 
+When merging a pull request to `master`, the squashed commit message should follow the [Conventional Commits](https://www.conventionalcommits.org) specification. This enables us to automatically generate CHANGELOGs & determine a semantic version bump.
```
lpww: So the version bump happens automatically based on commit messages? Is that a lerna feature?

jackdclark: Yeah, see https://github.com/lerna/lerna/tree/master/commands/version#--conventional-commits

The same can be achieved without Lerna, using [semantic-release](https://github.com/semantic-release/semantic-release)

lpww: cool! can you just start using the conventional commits at the sametime as semantic release? or will it mess things up if you didn't do it from the start?

lpww: I want to set this up on react-browser-hooks too

jackdclark: With Lerna you can definitely start half way through and it will start to pick things up from that point. I haven't used `semantic-release` myself, but it does provide a [dry run](https://semantic-release.gitbook.io/semantic-release/usage/configuration#dryrun) option where it prints out what it would do, so it's worth giving it a go locally.

PR #132, opened 2019-03-13
Fix default loading stage for useManualQuery

aribouius:
### What does this PR do?

Calling `useManualQuery` currently returns a state with `loading` set to _true_, despite the request function not having been called.  This PR updates the initial state for manual queries to default to _false_.

Resolves https://github.com/nearform/graphql-hooks/issues/133.

### Checklist

### Notes
- My implementation simply adds another option field `isManual` to identify the type of request being made, which is quite similar to the `isMutation` field in terms of usage. An alternative (but potentially more restrictive) approach would be to replace the two options with something to indicate the request will not immediately get called, e.g. `isDeferred`.  Please let me know if such an approach would be preferred.

### DIFF
--- a/packages/graphql-hooks/src/index.js
+++ b/packages/graphql-hooks/src/index.js
@@ -4,7 +4,7 @@ import useClientRequest from './useClientRequest'
 import useQuery from './useQuery'
 
 const useManualQuery = (query, options) =>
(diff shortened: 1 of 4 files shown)


bmullan91:
@allcontributors please add @aribouius for code and test.

Issue #91, opened 2019-03-04
Consider adding typings


jackdclark:
### Description

TypeScript type definitions can be bundled with an npm package, even if the lib is not written in TypeScript

See https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

jackdclark:
Typescript declaration files have been added in the latest releases for each package, these are:

- [graphql-hooks-memcache v1.2.0](https://github.com/nearform/graphql-hooks/releases/tag/graphql-hooks-memcache%401.2.0)
- [graphql-hooks-ssr v1.1.0](https://github.com/nearform/graphql-hooks/releases/tag/graphql-hooks-ssr%401.1.0)
- [graphql-hooks v3.4.0](https://github.com/nearform/graphql-hooks/releases/tag/graphql-hooks%403.4.0)

PR #135, opened 2019-03-13
feat: add a TypeScript declaration file for graphql-hooks

jackdclark:
### What does this PR do?

Adds a [https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) for `graphql-hooks` that will be published to npm, so that the package works well with TypeScript.

### Related issues

#91 

### Checklist

- [x] I have checked the [contributing document](../blob/master/CONTRIBUTING.md)
- [x] I have added or updated any relevant documentation
- [x] I have added or updated any relevant tests

PR #137, opened 2019-03-13
feat: add a TypeScript declaration file for graphql-hooks-memcache

jackdclark:
### What does this PR do?

Adds a [https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) for `graphql-hooks-memcache` that will be published to npm, so that the package works well with TypeScript.

### Related issues

#91 

### Checklist

### DIFF
--- a/packages/graphql-hooks-memcache/index.d.ts
+++ b/packages/graphql-hooks-memcache/index.d.ts
@@ -1,0 +1,16 @@
+export = MemCacheFunction
+
+declare function MemCacheFunction(options?: {
+  size?: number
+  ttl?: number
+  initialState?: object
+}): MemCache
+
+interface MemCache {
+  get(keyObject: object): object
+  set(keyObject: object, data: object): void
+  delete(keyObject: object): void
+  clear(): void
+  keys(): void
+  getInitialState(): object
+}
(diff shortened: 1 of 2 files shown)


packages/graphql-hooks-memcache/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "lib/graphql-hooks-memcache.js",
   "module": "es/graphql-hooks-memcache.js",
   "unpkg": "dist/graphql-hooks-memcache.min.js",
+  "types": "index.d.ts",
```
Joezo: I *think* you'll have to add this file to the `files` array on line 14

jackdclark: Ah yes good spot, I did this in #135.

packages/graphql-hooks-memcache/index.d.ts
```diff
@@ -0,0 +1,16 @@
+export = MemCacheFunction
+
+declare function MemCacheFunction(options?: {
+  size?: number
+  ttl?: number
+  initialState?: object
+}): MemCache
+
+interface MemCache {
+  get(keyObject: object): object
```
bmullan91: Technically `keyObject` could be `any`, but `graphql-hooks` gives it an object 😆

Issue #150, opened 2019-03-20
Inconsistent results for changing query


kkogovsek:
### Package
`graphql-hooks`

### Environment

- `graphql-hooks` version: 3.4.0
- `react` version: 16.8.4
- Browser: Any

### Description

When a query changes quickly results are inconsistent. What's wrong is that the current query is not tracked inside useClientRequest.js resulting in a race between different queries and the last one will be stored in state. In my case having consistent timing of server that means if I click next multiple times, I will see loading state at first and then multiple UI updates. Also, the loading state is lost since there's some of the old data in the state which arrived after all loading actions have been dispatched.

### How to reproduce

I created a simple code sandbox where you can click next/prev multiple times and see the issue in action.
https://codesandbox.io/s/mj6m8yvv4p

### Suggested solution (optional)

If this is undesired behavior I would like to open a PR solving that.

bmullan91:
@kkogovsek Good catch - we'd love to see a PR if you're up for it 🤗

bmullan91:
@kkogovsek Thanks again - the fix is now available in `graphql-hooks@3.4.1` 🎉

PR #153, opened 2019-03-20
Fixed race condition bug

kkogovsek:
### What does this PR do?

It adds ref holding latest requested cache key and on promise resolution checks for a match before dispatching data. Responses are still cached, but loading and result stay consistent between changes.

I'm struggling with a unit test since I would need to create a new client which returned the second result earlier than first in order to test it, but I would need some directions on how to structure that. Should I create two contexts and two clients to cover that case?

### Related issues

#150 

### Checklist


bmullan91:
@kkogovsek You'll need to create a separate client - ignoring the one created in `beforeEach`. In the mocked client you can then mock the `client.request` function so that the first call has a long setTimeout and the second call resolves immediately.

Maybe something along these lines might work:

```js
// NOTE I don't think you can use mockResolvedValue - since we want to delay it.
const firstP = new Promise((resolve) => setTimeout(() => resolve({ data: 'data' }), 200))
const secondP = Promise.resolve({ data: 'data' })

client.request = jest.fn()
  .mockReturnValueOnce(firstP)
  .mockReturnValueOnce(secondP)
```

kkogovsek:
I added two tests checking both possible situations:
- If we change query we want previous results to be omitted from changing state
- If the previous query was resolved after a new one we want to omit state update

packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -103,6 +104,13 @@ function useClientRequest(query, initialOpts = {}) {
     }
 
     const revisedCacheKey = client.getCacheKey(revisedOperation, revisedOpts)
+    const stringifiedCacheKey = JSON.stringify(revisedCacheKey)
```
bmullan91: We don't need to stringify the cache key - if it's the same key it'll be the same object reference.

packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -103,6 +104,13 @@ function useClientRequest(query, initialOpts = {}) {
...
+
+    // NOTE: Up until this point all calls will be called in correct order
+    // but after network request there's possibility of race conditions so
+    // activeCacheKey will enshure state to be updatet only with last value
+    activeCacheKey.current = stringifiedCacheKey
```
bmullan91: This can then become:

```suggestion
    activeCacheKey.current = revisedCacheKey
```

packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -130,7 +138,7 @@ function useClientRequest(query, initialOpts = {}) {
         client.cache.set(revisedCacheKey, result)
       }
 
-      if (isMounted.current) {
+      if (isMounted.current && stringifiedCacheKey === activeCacheKey.current) {
```
bmullan91:
```suggestion
      if (isMounted.current && revisedCacheKey === activeCacheKey.current) {
```

packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -103,6 +104,13 @@ function useClientRequest(query, initialOpts = {}) {
     }
 
     const revisedCacheKey = client.getCacheKey(revisedOperation, revisedOpts)
+    const stringifiedCacheKey = JSON.stringify(revisedCacheKey)
+
+    // NOTE: Up until this point all calls will be called in correct order
+    // but after network request there's possibility of race conditions so
+    // activeCacheKey will enshure state to be updatet only with last value
```
bmullan91: Minor: let's shorten this slightly and leave a link to the issue, perhaps some like:

```
NOTE: There is a possibility of a race condition whereby 
the second query could finish before the first one, dispatching an old result 
see https://github.com/nearform/graphql-hooks/issues/150 
```

kkogovsek:
> @kkogovsek Thanks for the PR - the fix works for me 🙌
> 
> I tried out setting `activeCacheKey.current = revisedCacheKey` in a fork of your original sandbox - see https://codesandbox.io/s/lyv5865w29

Yes I did the same 👍 thanks for taking your time reviewing my PR and awesome project

bmullan91:
@allcontributors[bot] add @kkogovsek for code and test

Issue #197, opened 2019-04-06
Show warning when using gql or passing in non-strings


good-idea:
### Description

I was migrating from Apollo and didn't remove `gql` before my query strings. The API I was querying returned some cryptic responses and it took a while to figure out that `gql` was causing the problem. The docs for graphql-hooks say "No need to wrap the query in `gql`", but it's easy to miss.

### Suggested implementation

Perhaps showing a warning in the console telling the developer to pass the query as a plain string? And, maybe notes in the docs that highlight that you should *not* use `gql` instead of "don't need to".

I'm happy to open a PR for this, just let me know how you'd like it to go!

bmullan91:
@good-idea Good idea 😜 A PR would be much appreciated!

good-idea:
No problem. How would you like to have it go? Just warn when it's not a string?

bmullan91:
Fixed in #199 and available in `graphql-hooks@3.4.2`

dohomi:
@bmullan91 would it not be better to use `print` function of `graphql/language/printer` to parse `gql` in case GQL is passed instead of a string?

PR #199, opened 2019-04-06
String warning

good-idea:
### What does this PR do?

Warns the user if they are passing in a query that is not a string. A common case might be if users use the `gql` template literal from graphql-tag on their queries.

### Related issues

#197 

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/useClientRequest.js
+++ b/packages/graphql-hooks/src/useClientRequest.js
@@ -51,6 +51,13 @@ function reducer(state, action) {
   opts.skipCache: Boolean
 */
 function useClientRequest(query, initialOpts = {}) {
+  if (typeof query !== 'string') {
+    /* eslint-disable-next-line no-console */
+    console.warn(
+      'Your query must be a string. If you are using the `gql` template literal from graphql-tag, remove it from your query.'
+    )
+  }
+
   const client = React.useContext(ClientContext)
   const isMounted = React.useRef(true)
   const activeCacheKey = React.useRef(null)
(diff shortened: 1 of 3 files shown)


good-idea:
Note: The test I wrote is working, but when trying to test this out on a local package with `npm link`, I got this hooks warning:

image

Any tips on how I should be setting this up?

good-idea:
@allcontributors[bot] please add @good-idea for query string checking

good-idea:
@allcontributors[bot] please add @good-idea for bug, code, test

packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -51,6 +51,13 @@ function reducer(state, action) {
   opts.skipCache: Boolean
 */
 function useClientRequest(query, initialOpts = {}) {
+  if (typeof query !== 'string') {
+    /* eslint-disable-next-line no-console */
+    console.warn(
```
bmullan91: @good-idea Since this will fail, we should throw an error.

Your warning message is great, let's use it for the error message - perhaps add at the end add `remove it from your query/mutation.` just in case the developer is confused if the error comes from `useMutation`.

bmullan91:
@good-idea RE https://github.com/nearform/graphql-hooks/pull/199#issuecomment-480528971:

1. run `npm run build` inside `packages/graphql-hooks`
1. use `npm pack` instead of `npm link`
1. install the tarball created from `npm pack` directly in your other repo to verify.

PR #175, opened 2019-03-31
Fix type declarations for logErrorResult & onError

wescoder:
The following throws without the propper type declarations when using `graphql-hooks` with typescript:
```
ERROR in [at-loader] ./node_modules/graphql-hooks/index.d.ts:44:25
    TS7031: Binding element 'Operation' implicitly has an 'any' type.

ERROR in [at-loader] ./node_modules/graphql-hooks/index.d.ts:44:44
    TS7031: Binding element 'Result' implicitly has an 'any' type.
```

### What does this PR do?

Fixes type errors when using typescript.

### Checklist

- [x] I have checked the [contributing document](../blob/master/CONTRIBUTING.md)
- [x] I have added or updated any relevant documentation
- [x] I have added or updated any relevant tests

PR #202, opened 2019-04-08
feat: add optional type annotations

good-idea:
### What does this PR do?

Adds the option to type the `data` from a request's response. Plus a small fix, see below.

Example:

```ts
const CURRENT_USER_QUERY = /*GraphQL*/`
	query UserQuery {
		user {
			id
			name
		}
	}
`

interface UserResponse {
	user: {
		id: string,
		username: string
	}
}

const User = () => {
	const { loading, error, data } = useQuery(CURRENT_USER_QUERY)
	if (!loading || !error) return null
	return <p>Hi, {data.user.name}</p>
}

const TypedUser = () => {
	const { loading, error, data } = useQuery<UserResponse>(CURRENT_USER_QUERY)
	if (!loading || !error) return null
	return <p>Hi, {data.user.name}</p> 
	/*                       👆
	 Property 'name' does not exist on type '{ id: string; username: string; }'
	*/
}
```

image

The types fall back to `any`, so it's backwards-compatible.

(I'm somewhat new to both typescript, this should get a review by someone who knows it better) 

Small fix: `graphQLError` should be `graphQLErrors`. (Right?)

https://github.com/nearform/graphql-hooks/blob/b8852a66f1555defe5f7413979a2f1e2e2e85df6/packages/graphql-hooks/index.d.ts#L121

### Checklist


packages/graphql-hooks/index.d.ts
```diff
@@ -56,7 +62,7 @@ interface ClientOptions {
   fetch?(url: string, options?: object): Promise<object>
   fetchOptions?: object
   logErrors?: boolean
-  onError?({ result, operation }: { operation: Operation, result: Result }): void
+  onError?({ operation: Operation, result: Result }): void
```
leonardodino: this is actually a regression of #175
please revert it.

```suggestion
  onError?({ result, operation }: { operation: Operation, result: Result }): void
```

packages/graphql-hooks/index.d.ts
```diff
@@ -105,23 +111,26 @@ interface UseQueryOptions extends UseClientRequestOptions {
   ssr?: boolean
 }
 
-interface UseClientRequestResult {
+interface UseClientRequestResult<Response> {
```
bmullan91: @good-idea Any reason we're using `ResponseData` as the generic name in other places, but `Response` here?

good-idea: Good catch, thanks!

packages/graphql-hooks/index.d.ts
good-idea: Oops! I think I had pasted this in from an out of date version. Prettier changed the formatting so I missed the diff.

package.json
```diff
@@ -9,7 +9,7 @@
     "postinstall": "lerna bootstrap --no-ci",
     "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
     "eslint": "eslint '**/*.js'",
-    "prettier": "prettier '**/*.js' '**/*.md' --write",
+    "prettier": "prettier '**/*.js' '**/*.md' '**/*.ts' --write",
```
bmullan91: Nice one @good-idea

bmullan91:
@good-idea This has been released `graphql-hooks@3.5.0` 🎉

Issue #234, opened 2019-05-13
Function returned from useMutation and useManualQuery is not memoized 


lynxtaa:
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: 3.5.0
- `react` version: 16.9.0-alpha.0
- Browser: Chrome

### Description

Function returned from useMutation and useManualQuery is not memoized and can't  be used in useEffect as a  dependency.

### How to reproduce

```jsx
const [shouldFetch, setShouldFetch] = useState(false)
const [fetchUser] = useManualQuery(GET_USER_QUERY)

useEffect(() => {
  if (shouldFetch) {
    fetchUser()
  }
}, [fetchUser, shouldFetch])

return <button onClick={() => setShouldFetch(true)} />
``` 

When button is clicked we expect only one request. But because fetchUser is not memoized internally, we get lots of requests and infinite render.

### Suggested solution (optional)

We can wrap fetchData in useCallback (see PR)

stagedataloss:
test123test123

lynxtaa:
As i found out you can use `client.request` from `ClientContext` (which doesn't change during rerenders) for manual queries and mutations in effects (Apollo has [similar approach](https://www.apollographql.com/docs/react/essentials/queries/#manually-firing-a-query)). Note, that caching won't work with `client.request`.

Example:
```jsx
import {ClientContext} from 'graphql-hooks'
...
const client = useContext(ClientContext)

useEffect() => {
  client.request({ query, variables })
}, [client])
```

lynxtaa:
I think to achieve memoization without deep quallity checks of `options` object (which also may contain functions like `fetchOptionsOverrides` and `updateData`) we should omit the second argument in useMutation and useManualQuery. Then invalidating returned function based on query will be easy. Although it'll be breaking change. @bmullan91 what do you think? I can create PR based on this idea

olistic:
@lynxtaa I find your example a bit contrived, to be honest. Is there a reason why you're using `useEffect` instead of calling `fetchUser` inside the `onClick` handler?

```js
const [fetchUser] = useManualQuery(GET_USER_QUERY)

return <button onClick={fetchUser} />
```

lynxtaa:
@olistic sorry for the confusion, I oversimplified my example to highlight main issue. Problem is: we can't use useManualQuery and useMutation inside useEffect because it creates infinite render. Here are two more examples:

Using `useManualQuery`:
```jsx
[...]

PR #235, opened 2019-05-13
fix: memoize callback for fetchData

lynxtaa:
### What does this PR do?

Memoize returned function from useMutation and useManualQuery so it can be used in useEffect as a  dependency. Consider this example:

```jsx

useEffect(() => {
  if (shouldFetch) {
    fetchUser()
  }
}, [fetchUser, shouldFetch])

``` 

### Related issues

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/useClientRequest.js
+++ b/packages/graphql-hooks/src/useClientRequest.js
@@ -103,71 +103,70 @@ function useClientRequest(query, initialOpts = {}) {
   latestStateData.current = state.data
 
   // arguments to fetchData override the useClientRequest arguments
-  const fetchData = React.useCallback(
-    newOpts => {
-      if (!isMounted.current) return Promise.resolve()
-      const revisedOpts = {
-        ...latestInitialOpts.current,
-        ...newOpts
-      }
+  function fetchData(newOpts) {
+    if (!isMounted.current) return Promise.resolve()
+    const revisedOpts = {
+      ...latestInitialOpts.current,
+      ...newOpts
+    }
 
-      const revisedOperation = {
-        ...operation.current,
-        variables: revisedOpts.variables,
-        operationName: revisedOpts.operationName
-      }
+    const revisedOperation = {
+      ...operation.current,
+      variables: revisedOpts.variables,
+      operationName: revisedOpts.operationName
+    }
 
-      const revisedCacheKey = client.getCacheKey(revisedOperation, revisedOpts)
+    const revisedCacheKey = client.getCacheKey(revisedOperation, revisedOpts)
 
-      // NOTE: There is a possibility of a race condition whereby
-      // the second query could finish before the first one, dispatching an old result
-      // see https://github.com/nearform/graphql-hooks/issues/150
-      activeCacheKey.current = revisedCacheKey
+    // NOTE: There is a possibility of a race condition whereby
+    // the second query could finish before the first one, dispatching an old result
+    // see https://github.com/nearform/graphql-hooks/issues/150
+    activeCacheKey.current = revisedCacheKey
 
-      const cacheHit =
-        revisedOpts.skipCache || !client.cache
-          ? null
-          : client.cache.get(revisedCacheKey)
+    const cacheHit =
+      revisedOpts.skipCache || !client.cache
+        ? null
+        : client.cache.get(revisedCacheKey)
 
-      if (cacheHit) {
-        dispatch({
-          type: actionTypes.CACHE_HIT,
-          result: cacheHit
-        })
-
-        return Promise.resolve(cacheHit)
-      }
-
-      dispatch({ type: actionTypes.LOADING })
-      return client.request(revisedOperation, revisedOpts).then(result => {
-        if (latestStateData.current && result.data && revisedOpts.updateData) {
-          if (typeof revisedOpts.updateData !== 'function') {
-            throw new Error('options.updateData must be a function')
-          }
-          result.data = revisedOpts.updateData(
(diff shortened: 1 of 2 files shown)


packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -60,13 +60,14 @@ function useClientRequest(query, initialOpts = {}) {
   const client = React.useContext(ClientContext)
   const isMounted = React.useRef(true)
   const activeCacheKey = React.useRef(null)
-  const operation = {
+  const operation = React.useRef(null)
+  operation.current = {
```
bmullan91: Stupid question (still getting my head around useRef): Why not pass the object as the initial value, instead of `null`?

packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -95,64 +96,76 @@ function useClientRequest(query, initialOpts = {}) {
...
+  latestStateData.current = state.data
+
   // arguments to fetchData override the useClientRequest arguments
-  function fetchData(newOpts) {
-    if (!isMounted.current) return Promise.resolve()
-    const revisedOpts = {
-      ...initialOpts,
-      ...newOpts
-    }
+  const fetchData = React.useCallback(
```
bmullan91: I think it might be clearer to leave this as-is, but at the end before we return the tuple we do the `useCallback` then:

```js
function fetchData() {...}

const memoisedFetchData = useCallback(fetchData, [client])

return [memoisedFetchData, state]
```

packages/graphql-hooks/test/unit/useClientRequest.test.js
```diff
@@ -441,6 +441,22 @@ describe('useClientRequest', () => {
...
+        },
+        { wrapper: Wrapper }
+      )
+
+      rerender()
+
+      expect(typeof fetchDataArr[0]).toBe('function')
+      expect(fetchDataArr[0]).toBe(fetchDataArr[1])
+    })
+
```
bmullan91: Could we include another test that verifies that the `fetchData` changes when each of the parameters change:
- `query` changes
- `options` changes

packages/graphql-hooks/src/useClientRequest.js
lynxtaa: By assigning `operation` on each render we make sure that inside `useCallback` we always have the latest version of `operation`.

`const operation = React.useRef({ ... })` will capture `query` and `initialOpts` only from the first render and may introduce bugs if these variables are changing between renders

packages/graphql-hooks/test/unit/useClientRequest.test.js
lynxtaa: `fetchData` is not changing when `options` changes because `options` is always new object.

Also `fetchData` is not changing when `query` changes. Consider this example:
```jsx
const [a, setA] = useState(true)
const [fetchUser] = useManualQuery(a ? GET_A : GET_B)

useEffect(() => {
  if (shouldFetch) {
    fetchUser()
  }
}, [fetchUser, shouldFetch])

return (
  <>
    <input type="checkbox" onChange={event => setA(event.target.checked)} checked={a} />
    <button onClick={() => setShouldFetch(true)} />
  </>
)
```

Changing checkbox value **should not** rerun the effect.

So, in this PR `fetchData` practically never changes. It captures `query` and `options` from the latest render.

packages/graphql-hooks/src/useClientRequest.js
lynxtaa: Done

lynxtaa: Although, this refactoring may be dangerous in long term because, as I found out, eslint-plugin-react-hooks will ignore this function. 

```jsx
function fetchData() {...}
const memoisedFetchData = useCallback(fetchData) // it's a mistake, but react-hooks linter rule is silent
```

```jsx
const fetchData = useCallback(() => {}) // it's a mistake and we get react-hooks/exhaustive-deps warning
```

packages/graphql-hooks/test/unit/useClientRequest.test.js
lynxtaa: Actually we can destructure operation and query from parsing `stringifiedCacheKey` and use it as a useCallback dependency. Then we don't need refs:

```jsx
function fetchData(...) {
  const { fetchOptions, operation } = JSON.parse(stringifiedCacheKey)
  ...
}

const memoisedFetchData = useCallback(fetchData, [client, stringifiedCacheKey])
```
It will change `fetchData` when query or options are changing. But i'm not sure if it's desirable (reruning the effects may be dangerous using useMutation hook). What do you think?

bmullan91: @lynxtaa sorry for the slow responses - super busy atm. To me, this makes more sense that the function changes when the query / options changes.

lynxtaa: After some experiments I found no way to be sure if passed options are really changed: it's an object which also can contain `updateData(...)` function. So naive approach with `JSON.stringify` won't work.

But I agree that function should change if query/options changes. And my current approach may introduce bugs. I'll think it through. 

Now I'm using `client.request` from `ClientContext` (which doesn't change during rerenders) for manual queries and mutations in effects (Apollo has [similar approach](https://www.apollographql.com/docs/react/essentials/queries/#manually-firing-a-query)).

lynxtaa:
Closing in favor of https://github.com/nearform/graphql-hooks/pull/285

PR #264, opened 2019-06-04
chore(CI): adds queuing step in acceptance-tests job

edvinasbartkus:
### What does this PR do?
The free version of browserstack.com is giving a limited number of free runners. If we have multiple PRs pending from renovatebot we initiate acceptance tests runners at the same time. That causes the job to fail since we run out of the browserstack runners.

With CircleCI queue orb we can pause the step and wait for other jobs to complete. As a result, we should not have acceptance-tests running concurrently.

The downside of this is that the execution time for every job might be very different.

### Related issues

N/A

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,8 @@
 # Check https://circleci.com/docs/2.0/language-javascript/ for more details
 #
 version: 2
+orbs:
+  queue: eddiewebb/queue@1.1.2
 jobs:
   build:
     working_directory: ~/graphql-hooks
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -3,6 +3,8 @@
 # Check https://circleci.com/docs/2.0/language-javascript/ for more details
 #
 version: 2
+orbs:
+  queue: eddiewebb/queue@volatile
```
Joezo: the `volatile` version is a bit scary. can we use the latest version [1.1.2](https://circleci.com/orbs/registry/orb/eddiewebb/queue)?

edvinasbartkus: :+1: updated to use 1.1.2

edvinasbartkus:
Also, added configuration option `consider-branch: false`, so the blocking wouldn't happen only for the same branches.

PR #266, opened 2019-06-05
feat: Add support for GraphQL multipart requests

### LINKED ISSUE #261: Add support for GraphQL multipart requests
### Description

Add support for GraphQL multipart requests following the [spec](https://github.com/jaydenseric/graphql-multipart-request-spec). This would allow to upload files with a GraphQL mutation.

### Suggested implementation

Use the [extract-files](https://github.com/jaydenseric/extract-files) package and follow an approach similar to [this](https://github.com/jaydenseric/graphql-react/blob/dfe7883850cb07e2a3f3ca38280db77e1d6f364d/src/universal/graphqlFetchOptions.mjs) in [here](https://github.com/nearform/graphql-hooks/blob/df6ee62cd7a81c079ec3d8f688514b7a287b8372/packages/graphql-hooks/src/GraphQLClient.js#L103).

Let me know what you think and I'll get started!

olistic:
### What does this PR do?

Allows to upload files via GraphQL multipart requests (see [spec](https://github.com/jaydenseric/graphql-multipart-request-spec)).

### Related issues

Closes #261 

### Checklist


olistic:
@jaydenseric Hi! I'm trying to add support for GraphQL multipart requests following your spec and using your `extract-files` library, however, [this](https://github.com/nearform/graphql-hooks/pull/266/files#diff-cd7376323827e8ded97f0163ffb000f2R315) assertion is failing in my tests. I also based my implementation in the `graphql-react`'s one, and [the same assertion doesn't fail for you](https://github.com/jaydenseric/graphql-react/blob/dfe7883850cb07e2a3f3ca38280db77e1d6f364d/src/test/graphqlFetchOptions.mjs#L42-L59). Do you have any idea why in my case body is not returning that Array shape but a more complex object?

cc @octet-stream I'm also using your `formdata-node` library, maybe you can shed some light as well.

Thanks to both of you! 🙏

octet-stream:
I'm not quite sure, but maybe it happens because node-fetch (which is used by jest-fetch-mock according to their dependencies) does not support formdata-node out of the box ([yet](https://github.com/bitinn/node-fetch/pull/603)), which means your `fetch` instance can't handle fields inside of the form-data properly.
To fix that, you may need to set `form.stream` instead of `form` itself as a body and set headers manually using `form.headers` property (which is just return an object with content-type header).

You also need to compare request body with the Readable stream rather than with FormData.

Hope it helps.

jaydenseric:
The [`deepEquals` assertion](https://github.com/tapjs/node-tap/blob/master/docs/asserts/index.md#tsamefound-wanted-message-extra) I'm using in `graphql-react` tests comes from [`tap`](https://npm.im/tap), which uses [`tcompare`](https://npm.im/tcompare). I think that works differently to your use of `toEqual` with [`jest`](https://npm.im/jest).

olistic:
@jaydenseric Yeah, that's exactly the issue. I had read the docs for `deepEquals` but didn't imagined the comparison algorithm was that loose.

@octet-stream Is there a way I can transform a `FormData` instance like this:

```js
const form = new FormData()
form.append('foo', 42)
```

into:

```js
[
  ['foo', 42]
]
```

so I can use `jest`'s `toEqual`?

olistic:
@bmullan91 This is ready for review now!

README.md
```diff
@@ -458,6 +460,12 @@ export default function PostList() {
 }
 ```
 
+## File uploads
+
+`graphql-hooks` complies with the [GraphQL multipart request spec](https://github.com/jaydenseric/graphql-multipart-request-spec), allowing files to be used as query or mutation arguments. The same spec is also supported by popular GraphQL servers, including [Apollo Server](https://www.apollographql.com/docs/apollo-server) (see list of supported servers [here](https://github.com/jaydenseric/graphql-multipart-request-spec#server)).
+
+If there are files to upload, the request's body will be a [`FormData`](https://developer.mozilla.org/docs/Web/API/FormData) instance conforming to the GraphQL multipart request spec.
+
```
bmullan91: @olistic Do you mind adding a concise code example here - perhaps the code from the tests?

packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -99,21 +101,53 @@ class GraphQLClient {
     }
   }
 
-  request(operation, options = {}) {
-    return this.fetch(this.url, {
+  getFetchOptions(operation, fetchOptionsOverrides = {}) {
```
bmullan91: Let's give kudos to @jaydenseric and link to the original source of this https://github.com/jaydenseric/graphql-react/blob/master/src/universal/graphqlFetchOptions.mjs 🙌

olistic:
@bmullan91 I addressed your comments in two separate commits. Please let me know what you think and if everything is good I'll squash them.

package.json
```diff
@@ -12,6 +12,9 @@
     "prettier": "prettier '**/*.js' '**/*.md' '**/*.ts' --write",
     "release": "lerna publish"
   },
+  "dependencies": {
+    "extract-files": "^5.0.1"
```
olistic: @bmullan91 I'm now wondering if I should have installed this in `graphql-hooks` specifically...

olistic: Addressed in d96baff.

edvinasbartkus:
LGTM :+1:

olistic:
@edvinasbartkus I squashed the `fixup!` commits and rebased against `master`, which dismissed your review 😬

edvinasbartkus:
@olistic reapproved :+1:

olistic:
@bmullan91 This is getting out-of-date pretty frequently. Let me know when you're ready to merge this and I'll rebase then.

kimbylr:
That's great! :) When are you planning on releasing this?

edvinasbartkus:
@kimbylr this has been released and should be available 3.6.0
Let me know if it does not work as expected.

PR #267, opened 2019-06-05
graphql-hooks-ssr: Fix incorrect type for React elements in TypeScript definitions file

tcudok-jg:
### What does this PR do?

Both `App` and `render` properties in `Options` were using an incorrect `ElementType` type. Trying to use the code from the README would result in a TypeScript file would result in an error:

```
Type 'Element' is not assignable to type 'ElementType<any>'.
  Type 'Element' is not assignable to type 'FunctionComponent<any>'. 

ts(2322)
```

Looking at React's type definitions, the type used for the element type in `ReactDOMServer.renderToStaticMarkup` is `ReactElement`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e253d335c1f036e67ace2898b598a3f620a6bb78/types/react-dom/server/index.d.ts#L36

This PR simply changes `ElementType` to `ReactElement` for both properties.

### Checklist


bmullan91:
@all-contributors please add @tcudok-jg  for code

PR #284, opened 2019-06-08
Memoise fetchData callback

### LINKED ISSUE #234: Function returned from useMutation and useManualQuery is not memoized

olistic:
### What does this PR do?

Memoises `fetchData` callback (function returned by `useMutation` and `useManualQuery`).

This is based on the work of @lynxtaa in #235 (who I credited in the commit), but uses a different strategy. In essence, it does a deep compare of the dependencies to `useCallback` between renders, to determine if they really changed. This is necessary because some of those dependencies are not primitive values but objects.

### Related issues

Fixes #234.

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/useClientRequest.js
+++ b/packages/graphql-hooks/src/useClientRequest.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import deepEqual from 'dequal'
 import ClientContext from './ClientContext'
 
 const actionTypes = {
@@ -40,6 +41,16 @@ function reducer(state, action) {
     default:
       return state
   }
+}
+
+function useDeepCompareCallback(callback, deps) {
+  const ref = React.useRef()
+
+  if (!deepEqual(deps, ref.current)) {
+    ref.current = deps
+  }
+
+  return React.useCallback(callback, ref.current)
 }
 
 /*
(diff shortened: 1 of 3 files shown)


packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -96,63 +107,66 @@ function useClientRequest(query, initialOpts = {}) {
...
+          })
+        }
 
-      return result
-    })
-  }
+        return result
+      })
+    },
+    [client, initialOpts, operation, state.data]
```
lynxtaa: Good Idea!

Correct me if I'm wrong but is `state.data` changing after successful fetch?  If so deep equal check will fail and effect in https://github.com/nearform/graphql-hooks/issues/234 example will rerun.

olistic: Yes, `state.data` will change, and it will cause `fetchData` to be re-created, but that is OK. `state.data` is a dependency of `fetchData` and it would be a mistake not to include it in `useCallback`'s dependencies. Regarding the example in #234, I think re-fetching could be avoided if you did things different. See my [comment](https://github.com/nearform/graphql-hooks/issues/234#issuecomment-500157999).

lynxtaa: So if it doesn't resolve the https://github.com/nearform/graphql-hooks/issues/234 , what problem it solves? Can you please please provide an example code where your memoization approach is beneficial? Thank you!

edvinasbartkus: Both of your arguments are valid that state is needed as a dependency and it does not fully solve the issue.

I have a suggestion here, let's remove the use of `state` from `fetchData`. We could pass `updateData` to reducer where state and `result.data` are available. That would eliminate the usage of state from `fetchData`.

For reducer:
```js
    case actionTypes.REQUEST_RESULT:
      return {
        ...action.result,
        data:
          action.updateData && state.data && action.result.data
            ? action.updateData(state.data, action.result.data)
            : state.data || action.result.data,
        cacheHit: false,
        loading: false
      }
```

in `fetchData` leaving the following bits:
```js
        if (
          revisedOpts.updateData &&
          typeof revisedOpts.updateData !== 'function'
        ) {
          throw new Error('options.updateData must be a function')
        }
```

```js
        if (isMounted.current && revisedCacheKey === activeCacheKey.current) {
          dispatch({
            type: actionTypes.REQUEST_RESULT,
            result,
            updateData: revisedOpts.updateData
          })
        }
```

@lynxtaa @olistic @bmullan91 what do you guys think?

lynxtaa: @edvinasbartkus Later we're storing updated result in cache:
```jsx
        result.data = revisedOpts.updateData(state.data, result.data)
      }

      if (revisedOpts.useCache && client.cache) {
        client.cache.set(revisedCacheKey, result)
      }
```

I'm not sure how it can be done with your approach.

The other way is to store the latest `state.data` in ref (like I do in https://github.com/nearform/graphql-hooks/pull/285).

ShogunPanda:
Superseded by #298. Closing it.

PR #285, opened 2019-06-08
fix: memoize useManualQuery and useMutation if no options passed

lynxtaa:
### What does this PR do?

If no initial options are passed, memoize function returned from useMutation and useManualQuery. As a result this function can be used in effects without causing infinite rerenders. Consider this example:

```jsx

useEffect(() => {
  if (shouldFetch) {
    fetchUser()
  }
}, [fetchUser, shouldFetch])

``` 

**Why proposed solution only works when no initial options are passed?**  I think this solution is the most straightforward. Memoization is tricky  and requires deep equality checks (also options may contain functions like `fetchOptionsOverrides` and `updateData`). And we can always pass these options when calling a memoized function. 

### Related issues

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/index.js
+++ b/packages/graphql-hooks/src/index.js
@@ -1,13 +1,20 @@
+import { useMemo } from 'react'
 import ClientContext from './ClientContext'
 import GraphQLClient from './GraphQLClient'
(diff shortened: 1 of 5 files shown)


ShogunPanda:
Superseded by #298. Closing it.

PR #298, opened 2019-06-13
Memoise fetch data cache

edvinasbartkus:
### What does this PR do?

It supplements @olistic's and @lynxtaa's ideas about memoisation by moving the caching into the client, making it possible to avoid references of state in fetchData function.

### Related issues

https://github.com/nearform/graphql-hooks/pull/284
https://github.com/nearform/graphql-hooks/pull/285

### Checklist

- [ ] I have checked the [contributing document](../blob/master/CONTRIBUTING.md)


packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -20,28 +21,41 @@ function reducer(state, action) {
...
+    return currentData
+  }
+
+  return updater(currentData, newData)
+}
+
+function useDeepCompareCallback(callback, deps) {
+  const ref = React.useRef()
+
+  if (!deepEqual(deps, ref.current)) {
```
mcollina: Isn't this `deepEqual` check going to be very expensive?

packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react'
+import deepEqual from 'dequal'
```
mcollina: This has very few downloads/month. Is there a more popular one? Can we use one from the lodash suite for example?

olistic: @mcollina I chose it because it does what it has to do, plus:

- It's well tested.
- It's tiny ([347B](https://bundlephobia.com/result?p=dequal@1.0.0) vs [3.8kB](https://bundlephobia.com/result?p=lodash.isequal@4.5.0) lodash's `isEqual`).
- It's [faster](https://github.com/lukeed/dequal#benchmarks) than lodash's `isEqual` (3.54x faster, to be specific).

mcollina: Awesome!

packages/graphql-hooks/src/useClientRequest.js
olistic: I don't think so given what we are comparing. And unless we get rid of non-primitive dependencies in `useCallback` calls (`initialOpts`) I think we're stuck to something like this if we want to memoise the `fetchData` callback.

Also, it's possible that the `deepEqual` is cheaper than whatever the user is doing within any hooks that have a `fetchData` that is changing unnecessarily as a dependency.

edvinasbartkus: my goal with this PR (changes on top of @olistic suggestions) was to remove `state.data` from useCallback dependencies that can cause expensive comparison. The downside is that caching has to be moved to `GraphQLClient`. I think without `state` in dependency array the array itself is pretty lean: `variables`, `options: { useCache, skipCache, etc. }`.

packages/graphql-hooks/src/useClientRequest.js
lynxtaa: I noticed that `dequal` doesn't compare functions:
```jsx
const dequal = require('dequal')
dequal((a, b) => a + b,  (a, b) => a + b) // false
```
So using passing `fetchOptionsOverrides` and `updateData` to `options` will break memoization.

edvinasbartkus: @lynxtaa

But neither `dequal` nor another package is able to do that kind of function comparison. In the following example the memoization will happen properly using `useCallback`:

```js
const { useCallback } from 'react'
const updateData = useCallback((old, new) => {
  return new
}, [])

const { data, refetch } = useQuery(GQL, { updateData })
```

bmullan91:
Available in `graphql-hooks@3.6.1`

packages/graphql-hooks/src/useClientRequest.js
heymartinadams: You can also use `fast-deep-equal` instead of `dequal` — about a [33% increase](https://www.npmjs.com/package/dequal#benchmarks) in performance, although at this level I’m not sure it makes a difference.

PR #303, opened 2019-06-14
Updates renovate config: bumping only major versions

edvinasbartkus:
### What does this PR do?

Updates renovate bot config to only make PRs for major version changes. Also, it will not force the pin of the version when doing a PR.

https://renovatebot.com/docs/configuration-options/#bumpversion
https://renovatebot.com/docs/configuration-options/#rangestrategy

### Checklist


mcollina:
I would like to change https://github.com/nearform/graphql-hooks/blob/master/package.json so that it depends with `^`.

PR #310, opened 2019-06-19
Unpin

edvinasbartkus:
### What does this PR do?

Removing the configuration `bumpVersion`. Seems like it's addressing package.json#version attribute but we want to keep version bumps a manual process.

Unpins all versions in subpackages and example projects.

### Related issues

N/A

### Checklist


edvinasbartkus:
@mcollina to my understanding - yes, should be enough of `rangeStrategy: replace` to address that updates will happen only for major versions:
```
replace = Replace the range with a newer one
  if the new version falls outside it, e.g. ^1.0.0 -> ^2.0.0
```

Issue #313, opened 2019-06-25
Shouldn't refetch clear errors?


olistic:
### Description

I'm using `refetch` to retry a failed query, like this:

```js
const List = () => {
  const { loading, error, data, refetch } = useQuery(MY_LIST_QUERY)

  if (loading || error) {
    return <Loading error={error} retry={refetch} />
  }

  // ...
}

const Loading = ({ error, retry }) => {
  if (error) {
    return <button onClick={retry}>Retry</button>
  }

  return <Spinner />
}
```

I expect that when I click "Retry" (which calls `refetch`) `loading` is set back to `true` and any errors from previous fetch are cleared, so I see the spinner again.

Does this make sense?

olistic:
@edvinasbartkus Were you able to take a look at this? Thanks

bmullan91:
@olistic Sorry for the delay, yea I think you're right - we should clear the errors.

In the meantime, you can handle them as two separate states:

```js
if (loading) {
  return <Loading />
}

if (error) {
  return <Loading error={error} retry={refetch} />
}
```

I think you're overloading the `Loading` component here btw, might be worth having a separate component/logic for handling errors.

edvinasbartkus:
Sorry for the delay. Some vacation time got on a way. I have looked it up the changes I have prepared earlier and rebased with the current master:
https://github.com/nearform/graphql-hooks/pull/339

My suggestion is simply to check before going to loading state if there was an error before. If so, we can reset to the initial state + loading.

PR #339, opened 2019-07-29
fix: reset state with an error immediately on data refetch

edvinasbartkus:
### What does this PR do?

Once we go into the loading state we can check if the previous state includes any error, if so we reset it to the initial state.

### Related issues

Solves https://github.com/nearform/graphql-hooks/issues/313

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/useClientRequest.js
+++ b/packages/graphql-hooks/src/useClientRequest.js
@@ -14,6 +14,14 @@ function reducer(state, action) {
     case actionTypes.RESET_STATE:
       return action.initialState
     case actionTypes.LOADING:
+      // if previous action resulted in error - refetch will reset the state
+      if (state.error) {
+        return {
+          ...action.initialState,
+          data: state.data,
+          loading: true
+        }
+      }
       if (state.loading) {
         return state // saves a render cycle as state is the same
       }
@@ -145,7 +153,7 @@ function useClientRequest(query, initialOpts = {}) {
         return Promise.resolve(cacheHit)
       }
 
-      dispatch({ type: actionTypes.LOADING })
+      dispatch({ type: actionTypes.LOADING, initialState })
       return client.request(revisedOperation, revisedOpts).then(result => {
         if (
           revisedOpts.updateData &&
--- a/packages/graphql-hooks/test/unit/useClientRequest.test.js
+++ b/packages/graphql-hooks/test/unit/useClientRequest.test.js
@@ -77,6 +77,45 @@ describe('useClientRequest', () => {
     })
   })
 
+  it('does reset data if previous call resulted in error', async () => {
+    let fetchData
+    let state
+
+    renderHook(
+      () =>
+        ([fetchData, state] = useClientRequest(TEST_QUERY, {
+          updateData: () => {}
+        })),
+      {
+        wrapper: Wrapper
+      }
+    )
+
+    mockClient.request.mockResolvedValueOnce({
+      error: true,
+      errors: ['oh no!']
+    })
+
+    await fetchData()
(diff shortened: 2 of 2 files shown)


packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -14,6 +14,13 @@ function reducer(state, action) {
     case actionTypes.RESET_STATE:
       return action.initialState
     case actionTypes.LOADING:
+      // if previous action resulted in error - refetch will reset the state
+      if (state.error) {
+        return {
+          ...action.initialState,
```
bmullan91: @edvinasbartkus this will clear the `state.data` which could be set from a previous call - we should only clear the error and set loading to true.

packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -145,7 +152,7 @@ function useClientRequest(query, initialOpts = {}) {
         return Promise.resolve(cacheHit)
       }
 
-      dispatch({ type: actionTypes.LOADING })
+      dispatch({ type: actionTypes.LOADING, initialState })
```
bmullan91: No need to pass initialState here - see other comment.

packages/graphql-hooks/src/useClientRequest.js
edvinasbartkus: but can we have a state where data is available and also error would be true?

Maybe the best would be only to persist `data`, something like that:
```js
return {
  ...action.initialState,
  data: state.data,
  loading: true
}
```

otherwise, I guess we need to do something like that to cleanup all error properties?
```js
return {
  ...state,
  error: false,
  fetchError: undefined,
  httpError: undefined,
  graphQLErrors: undefined
}
```

bmullan91: @edvinasbartkus this scenario can happen when calling `refetch`... Yea I'd probably go with the first option 👍

bmullan91:
@all-contributors please add @edvinasbartkus  for question bug doc example ideas maintenance review test

packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -14,6 +14,14 @@ function reducer(state, action) {
     case actionTypes.RESET_STATE:
       return action.initialState
     case actionTypes.LOADING:
+      // if previous action resulted in error - refetch will reset the state
```
bmullan91:
```suggestion
      // if the previous action resulted in an error - refetch should clear all errors
```

packages/graphql-hooks/test/unit/useClientRequest.test.js
```diff
@@ -77,6 +77,45 @@ describe('useClientRequest', () => {
     })
   })
 
+  it('does reset data if previous call resulted in error', async () => {
```
bmullan91:
```suggestion
  it('should clear errors when calling fetchData', async () => {
```

packages/graphql-hooks/test/unit/useClientRequest.test.js
```diff
@@ -77,6 +77,45 @@ describe('useClientRequest', () => {
...
+    const promise = new Promise(resolve => {
+      promiseResolve = resolve
+    })
+
+    mockClient.request.mockResolvedValueOnce(promise)
+    fetchData()
+
+    expect(state).toEqual({ cacheHit: false, loading: true })
+    promiseResolve()
+  })
```
bmullan91: @edvinasbartkus do mind adding one more test to verify that the data is maintained.

bmullan91:
closing in favour of https://github.com/nearform/graphql-hooks/pull/469

Issue #366, opened 2019-09-24
tiny_lru__WEBPACK_IMPORTED_MODULE_0___default


and3rsonsousa:
### Package

`graphql-hooks`
`graphql-hooks-ssr`
`graphql-hooks-memcache`

### Environment

- `graphql-hooks` version: 3.2.1
- `graphql-hooks-ssr` version: 1.0.1
- `graphql-hooks-memcache` version: 1.0.4
- `react` version:  16.8.2
- Browser: Chrome 77.0.3865.90

### Description

Started to get this error one day ago only on localhost, on now(zeit) works properly. Downloaded Next.js example using graphql-hooks, no modification on the code 
and the same error has happened.
<img width="1038" alt="Captura de Tela 2019-09-24 às 01 31 35" src="https://user-images.githubusercontent.com/29865417/65481299-1c596b80-de6b-11e9-9a35-fd0953957dad.png">

fabienheureux:
Exact same issue here 😶

PR #369, opened 2019-09-26
Swap tiny-lru with lru-cache, tiny-lru causing import error

fabienheureux:
### What does this PR do?

tiny-lru caused import error, even on a very basic example (cf nextjs/examples repo). 

Swapping tiny-lru by another lru package does the trick. 
tiny-lru does not seem very suitable for es6 import: 
- https://github.com/avoidwork/tiny-lru#example  
- https://github.com/avoidwork/tiny-lru/issues/19

### Related issues

https://github.com/nearform/graphql-hooks/issues/366

### Checklist

### DIFF
--- a/packages/graphql-hooks-memcache/.size-snapshot.json
+++ b/packages/graphql-hooks-memcache/.size-snapshot.json
@@ -1,46 +1,46 @@
 {
   "lib/graphql-hooks-memcache.js": {
-    "bundled": 936,
-    "minified": 621,
-    "gzipped": 369
+    "bundled": 989,
+    "minified": 666,
+    "gzipped": 396
   },
   "es/graphql-hooks-memcache.js": {
-    "bundled": 763,
-    "minified": 490,
-    "gzipped": 311,
+    "bundled": 816,
+    "minified": 535,
+    "gzipped": 339,
     "treeshaked": {
       "rollup": {
-        "code": 45,
-        "import_statements": 45
+        "code": 68,
+        "import_statements": 46
       },
       "webpack": {
-        "code": 1062
+        "code": 1085
       }
     }
   },
   "es/graphql-hooks-memcache.mjs": {
-    "bundled": 2080,
-    "minified": 2076,
-    "gzipped": 886,
+    "bundled": 1706,
+    "minified": 1706,
+    "gzipped": 725,
     "treeshaked": {
       "rollup": {
-        "code": 1471,
+        "code": 22,
         "import_statements": 0
       },
       "webpack": {
-        "code": 2521
+        "code": 1121
       }
     }
   },
   "dist/graphql-hooks-memcache.js": {
-    "bundled": 4451,
-    "minified": 2281,
-    "gzipped": 973
+    "bundled": 3427,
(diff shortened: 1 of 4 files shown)


packages/graphql-hooks-memcache/.size-snapshot.json
```diff
@@ -1,46 +1,46 @@
...
+        "code": 46,
+        "import_statements": 46
       },
       "webpack": {
-        "code": 1062
+        "code": 1063
       }
     }
   },
   "es/graphql-hooks-memcache.mjs": {
```
hboylan: This looks good. Any idea why there's ~4x increased bundle size for `es/` and `dist/`?

fabienheureux: It seems that lru-cache has one dependency whereas tiny-lru has none.
I used one of the most used lru package I found, but maybe you have a better candidate ?

hboylan: No, that looks like a good candidate. I just wonder if it's possible to continue using `tiny-lru`. [This section](https://www.npmjs.com/package/tiny-lru#example) of their README describes how it should be imported or aliased.

The `graphql-hooks-memcache` tests pass with `import LRU from 'tiny-lru/lib/tiny-lru.es5.js'`, but I'm unable to whether this fixes the Webpack import issue.

Another option might be to try and alias `tiny-lru` in the rollup config, which would require `rollup-plugin-alias`.

fabienheureux: I tried the method described (that's why I link it in the initial PR description), but with no luck unfortunately, the Webpack import issue persisted. 

I am not really used to rollup, and was not sure if alias would work. I will try using this plugin, IMO adding a dev dependency is better than multiply the bundled size by 4.

hboylan: Ah, sorry I overlooked that link.

Same here. Thanks for taking a look. I wasn't sure of the best way to test the build with webpack. 🤞

fabienheureux: If ever you want to reproduce / play with this bug, you can clone the next.js / with-graphql-hooks example here: https://github.com/zeit/next.js/tree/canary/examples/with-graphql-hooks
```
npm i && npm run dev
```
will show you the error

osdevisnot: I am not sure if the new options is most optimal. It does bloat the bundle size quite a bit.

osdevisnot: @fabienheureux @hboylan - let me know what you think of #370 

I am also trying to get `tiny-lru` updated here https://github.com/avoidwork/tiny-lru/pull/27

fabienheureux:
Closing in favor of https://github.com/nearform/graphql-hooks/pull/370

packages/graphql-hooks-memcache/.size-snapshot.json
fabienheureux: LGTM

Issue #72, opened 2019-02-28
Add support for Subscriptions


Joezo:
### Description

We support Queries and Mutations but don't support the third operation type, Subscriptions. We should look at adding that.

### Suggested implementation
I would guess websockets?

mcollina:
Subscriptions will be extraordinarly complicated. I would restrain from using (and implementing) them, and they are not really part of the spec (See https://github.com/facebook/graphql/blob/master/rfcs/Subscriptions.md, it's a draft).

On the server side, they would scale _horribly_. I've worked on pubsub systems, and this type of unbounded updates is extremely hard to use in production.

It would be fine to experiment with those, but as a separate module.

Joezo:
I think you're right @mcollina - it does look incredibly complicated. Might be worth keeping this around as a long running issue to track subscription discussion.

brainkim:
> On the server side, they would scale horribly. I've worked on pubsub systems, and this type of unbounded updates is extremely hard to use in production.

@mcollina Genuinely curious to hear more on this. What do you mean by “unbounded” updates? Are `apollographql/graphql-subscriptions` and packages which implement the `PubSubEngine` interface fatally flawed?

mcollina:
It closely depends on how you code the resolvers: it's extremely easy to shoot yourself in the foot.

The following example is essentially flowed:

```js
const SOMETHING_CHANGED_TOPIC = 'something_changed';

export const resolvers = {
  Subscription: {
    somethingChanged: {
      subscribe: () => withFilter(
        (_, args) => pubsub.asyncIterator(SOMETHING_CHANGED_TOPIC),
        (payload, variables) => payload.anAttribute === variables.AnAttribute,
      ),
  },
}
```

Essentially you are doing the vast majority of the filtering on your Node.js processes. This __will not scale__, because the processes will receive the vast majority of the updates. You need to use different topics and _never ever filter_ within a resolver. At some point, the filtering will use 100% of the CPU of every process without sending any data. Doing filtering within the resolvers (as all example do) works well at small scale but is fatally flawed on a huge volume of messages moved. 

The whole purpose of GraphQL is doing advanced filtering. Not be able to do filtering makes it a non compelling technology compared to alternatives for real-time communication.

good-idea:
What would make this difficult on the client?

brainkim:
@mcollina Had an aha moment today after reading your comment. To rephrase what you’re saying for my own understanding:

[...]

PR #394, opened 2019-10-30
useSubscription hook

frikille:
### What does this PR do?

Implementing useSubscription in graphql-hooks

This PR is more like a proof of concept at the moment after spending some time learning about different useSubscription implementation.

The current implementation is very simple and is intended to start a discussion on how the final implementation and API should look like. This implementation required the minimal code to get started and it doesn't add new dependencies to the graphql-hooks project.

The GraphQLClient has been modified and it is possible to set a new config option called `subscriptionClient` which must be an instance of `SubscriptionClient` from the (`subscription-transport-ws`)[https://github.com/apollographql/subscriptions-transport-ws] package. By using the `SubscriptionClient` there is no need to implement anything new apart of the `useSubscription` function in `graphql-hooks`. However, I think it's worth the discussion around using the SubscriptionClient from that package or implementing our own as that library doesn't seem to get too much care from its maintainers.

If we decide to implement our own library that probably introduces some new dependencies however that can be mitigated by creating a new package and pass the subscription client instance to the GraphQLClient during instantiation.

### Related issues

#72 

### Checklist


mcollina:
I would just go ahead and ship our own implementation instead of `subscription-transport-ws`.

examples/subscription/src/server/index.js
```diff
@@ -0,0 +1,128 @@
+const fastify = require('fastify')
+const GQL = require('fastify-gql')
+const mq = require('mqemitter')
+const emitter = mq()
```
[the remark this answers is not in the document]
frikille: Will update the example using Redis

examples/subscription/src/server/index.js
```diff
@@ -0,0 +1,133 @@
+const fastify = require('fastify')
```
mcollina: use strict is missing at the top of the file.

examples/subscription/webpack.config.js
```diff
@@ -0,0 +1,81 @@
+const path = require('path')
```
mcollina: use strict is missing

packages/graphql-hooks-subscription-client/test/index.test.js
```diff
@@ -0,0 +1,160 @@
...
+    server.close()
+    WS.clean()
+    server = new WS('ws://localhost:8888', { jsonProtocol: true })
+    server.send({ id: undefined, type: 'connection_ack' })
+
+    setTimeout(() => {
+      process.nextTick(() => {
+        server.send({ id: undefined, type: 'connection_ack' })
+      })
+    }, 1000)
```
mcollina: I would prefer if we could rewrite this test to not use setTimeout.

packages/graphql-hooks/test/unit/useSubscription.test.js
```diff
@@ -0,0 +1,229 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { renderHook } from '@testing-library/react-hooks'
+import { ClientContext, useSubscription, GraphQLClient } from '../../src'
+
+class MockSubscriptionClient {
```
mcollina: Why is this mock needed? Can you add some comments on why?

simoneb:
@frikille thanks for your work on this, it's amazing! I skimmed through the code quickly and it all makes sense to me. The main question I have to start a discussion around this is, where does the code of `SubscriptionClient` come from? I suppose it comes from `subscription-transport-ws` and you copied its code here? 

If so, we may need to review how we want to move this forward, in the sense that we have a few options:

1) we reference that library
2) we copy the code
3) we implement our own

I think 1) doesn't adhere to the principle that's being applied throughout this library of avoiding third party dependencies. 2) would require making sure that we are allowed to copy the code and give proper attribution (haven't checked the licensing), and 3) I think is the preferred option, as well as the one requiring more effort of course.

bmullan91:
@frikille thanks for your work here, as discussed the initial approach using `subscriptions-transport-ws` is ideal - we _really_ don't want to maintain our own flavour of that. I know you're busy right now, so I went ahead and cherry-picked the necessary commits over to #415 with a few tweaks.

Issue #337, opened 2019-07-26
onError doesn't do what it suggests


levrik:
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: 3.6.0

### Description

I wanted to use `onError` to implement a JWT session refresh after a failed request and then retry the request. But now I had to find out that `onError` only gets called if error logging is enabled (which already is quite weird). Also I can't return another `Promise` from `onError` to actually still resolve the request.

### Suggested solution

I could propose a solution to that via a PR if you are fine.
Now I have to wrap `axios` in some `fetch`-like API like suggested at the end of the README.
Actually I don't want to ship `axios` since it adds additional size to my bundle, also I don't want every request go through another layer of abstraction like the suggested fetch-wrapper.

bmullan91:
@levrik sorry for the delay; I agree `onError` isn't that useful for this. You can solve this by passing your own `fetch` implementation, something along these lines:

```js
function customFetch(url, options) {
  return fetch(url, options).catch(error => {
    if (error.message === 'expired jwt') {
      refreshJWT().then(() => {
        // retry...
        return fetch(url, options);
      })
    }

    // re-throw 
    throw error
  })
}

const client = new GraphQLClient({
  url: '/graphql',
  fetch: customFetch
})
```

levrik:
@bmullan91 Awesome. Thanks!

Issue #85, opened 2019-03-04
Console messages are nested


Haroenv:
### Environment

- `graphql-hooks` version: 3.2.1
- `react` version: 16.8.3
- Browser: Safari & Chrome

### Description

Second errors are nested within the first error on refetch.

screenshots are Safari & Chrome

<img width="646" alt="screenshot 2019-03-04 at 11 16 58" src="https://user-images.githubusercontent.com/6270048/53727206-ed988b00-3e6f-11e9-91b8-7e186d99bf99.png">
<img width="629" alt="screenshot 2019-03-04 at 11 18 22" src="https://user-images.githubusercontent.com/6270048/53727207-ed988b00-3e6f-11e9-84f5-b7655df8a764.png">

### How to reproduce

Something which fails, e.g. https://codesandbox.io/s/42k89kzpp7 (use expanded form to get the full console which has the error)

### Suggested solution (optional)

Error should be somewhere here, but maybe just nested groups aren't possible? 

https://github.com/nearform/graphql-hooks/blob/ae154d3a0b144416f2fe5d168585c22e74ad68ba/src/GraphQLClient.js#L45

bmullan91:
We're missing a final `console.groupEnd()` 😅

bmullan91:
@Haroenv fixed in `v3.2.2`

Haroenv:
Works perfectly, thanks!

bmullan91:
@all-contributors please add @Haroenv for bug

PR #419, opened 2019-12-04
Fix onerror not called

panz3r:
### What does this PR do?

This PR decouples `onError` custom error handler from `logErrors` flag to allow the use of a custom handler without the need to set `logErrors` flag to `true`.
This align the actual behaviour of the 2 options to what is documented in the `README`.

### Related issues

#337 

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/GraphQLClient.js
+++ b/packages/graphql-hooks/src/GraphQLClient.js
@@ -50,38 +50,47 @@ class GraphQLClient {
   }
   /* eslint-disable no-console */
   logErrorResult({ result, operation }) {
-    if (this.onError) {
-      return this.onError({ result, operation })
-    }
+    if (this.logErrors) {
+      console.error('GraphQL Hooks Error')
+      console.groupCollapsed('---> Full Error Details')
+      console.groupCollapsed('Operation:')
+      console.log(operation)
+      console.groupEnd()
 
-    console.error('GraphQL Hooks Error')
-    console.groupCollapsed('---> Full Error Details')
-    console.groupCollapsed('Operation:')
-    console.log(operation)
-    console.groupEnd()
+      const error = result.error
 
-    const error = result.error
+      if (!error) {
+        return
+      }
 
-    if (error.fetchError) {
-      console.groupCollapsed('FETCH ERROR:')
-      console.log(error.fetchError)
+      if (error.fetchError) {
+        console.groupCollapsed('FETCH ERROR:')
+        console.log(error.fetchError)
+        console.groupEnd()
+      }
+
+      if (error.httpError) {
+        console.groupCollapsed('HTTP ERROR:')
+        console.log(error.httpError)
+        console.groupEnd()
+      }
+
+      if (error.graphQLErrors && error.graphQLErrors.length > 0) {
+        console.groupCollapsed('GRAPHQL ERROR:')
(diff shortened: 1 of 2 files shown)


packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -219,8 +228,9 @@ class GraphQLClient {
         })
       })
       .then(result => {
-        if (result.error && this.logErrors) {
+        if (result.error) {
           this.logErrorResult({ result, operation })
+          this.notifyErrorResult({ result, operation })
         }
```
bmullan91: Let's do the `if's` here, removing the responsibility from `logErrorResult` etc:

```js
if (result.error) {
  if (this.logErrors) {
    this.logErrorResult({ result, operation })
  }

  if (this.onError) {
    this.onError({ result, operation })
  }
}
```

packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -50,38 +50,47 @@ class GraphQLClient {
...
 
-    if (error.graphQLErrors && error.graphQLErrors.length > 0) {
-      console.groupCollapsed('GRAPHQL ERROR:')
-      error.graphQLErrors.forEach(err => console.log(err))
       console.groupEnd()
     }
+  }
 
-    console.groupEnd()
+  notifyErrorResult({ result, operation }) {
```
bmullan91: We can remove this and call `this.onError` in the `if (result.error)` block.

panz3r:
Hi @bmullan91 ,

I updated the PR code as requested and also the unit-tests to reflect the changes.

packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -62,6 +58,10 @@ class GraphQLClient {
 
     const error = result.error
 
+    if (!error) {
```
bmullan91: @panz3r If we return here, we'd have never called the final `console.groupEnd`, meaning the all further console logs will be nested. This happened before https://github.com/nearform/graphql-hooks/issues/85.

An alternative would be to add check in each `if ` below:

```js
if (error && error.fetchError) {...}
if (error && error.httpError) {...}
if (error && error.graphQLErrors && error.graphQLErrors.length > 0) {...}
```

packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -219,8 +220,14 @@ class GraphQLClient {
...
-        if (result.error && this.logErrors) {
-          this.logErrorResult({ result, operation })
+        if (result.error) {
+          if (this.logErrors) {
+            this.logErrorResult({ result, operation })
+          }
+
+          if (this.onError) {
+            this.onError({ result, operation })
+          }
```
bmullan91: Nice 👍

packages/graphql-hooks/src/GraphQLClient.js
panz3r: To avoid repeating the issue #85, I improved the tests around `logErrorResult` method and updated the skip implementation.

bmullan91:
@all-contributors please add @panz3r for code test

bmullan91:
@panz3r available in `graphql-hooks@4.3.0` 🚀

Issue #331, opened 2019-07-16
Issue in Edge Causing JS Parsing Error


joealden:
### Package

`graphql-hooks-memcache`

### Environment

- `graphql-hooks` version: `3.6.0`
- `graphql-hooks-memcache` version: `1.2.1`
- `react` version: `16.8.6`
- Browser: Edge 18

### Description

When visiting my site in Edge, I am getting the following error logged to the console:

```
SCRIPT1028: Expected identifier, string or number
```

From a little research, and from looking at where the error happened in the source-map (shown in the below screenshot), I think this is happening because [object spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals) is being used, and is not getting compiled down to code Edge can understand (Edge currently doesn't support the object spread syntax :cry:).

![screenshot](https://user-images.githubusercontent.com/10297243/61335814-bf1bbb00-a826-11e9-99c2-81f2014a2cf3.png)

### How to reproduce

Create a project that uses `graphql-hooks-memcache`, then try to load the site in Edge.

### Suggested solution

If my assumptions are correct about the cause of the issue, either compile the syntax down to syntax Edge understands, or change the source to not use the object spread syntax at all.

---

Thanks for your time!

mcollina:
How are you bundling/transpiling this? I fear it might just be some missing configuration on webpack/babel.

joealden:
@mcollina I am using a relatively new version of GatsbyJS, but a version before they started transpiling `node_modules` (I haven't upgraded due to a build issue we are having with our project with the latest version).

From your above comment, am I right in thinking that you expect people to configure webpack + babel to transpile their project's dependencies?

bmullan91:
@joealden Good catch! Seems we're missing this `.babelrc.js` in `graphql-hooks-memcache`, I'm hopefull using the same one from here should do the trick: https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks/src/.babelrc.js

@joealden I don't have easy access to a windows machine to test/debug, do you fancy trying that out?

Steps:
1. Read the `for developers` section in the contributing guide: https://github.com/nearform/graphql-hooks/blob/master/CONTRIBUTING.md#for-developers
1. Copy `babelrc.js` from `packages/graphql-hooks/src` into `packages/graphql-hooks-memcache`
1. `npm run build` inside `packages/graphql-hooks-memcache`
[...]

PR #438, opened 2020-01-10
fix: Transpile to ES5 for support in IE/Edge

WayneEllery:
### What does this PR do?

Fixes errors in IE and Edge for `graphql-hooks-memcache` by transpiling to ES5.

I copied the `.babelrc.js` from `graphql-hooks/src/` into the root directory of `graphql-hooks-memcache`. I think it's better to be in root but I can put in `src` if that's what you would prefer.

### Related issues

 #331

### Checklist

### DIFF
--- a/packages/graphql-hooks-memcache/src/.babelrc.js
+++ b/packages/graphql-hooks-memcache/src/.babelrc.js
@@ -1,0 +1,21 @@
+const { NODE_ENV } = process.env
+
+module.exports = {
+  presets: [
+    [
+      '@babel/env',
+      {
+        targets: {
+          browsers: ['ie >= 11']
+        },
+        modules: false,
+        loose: true
+      }
+    ]
+  ],
+  plugins: [
+    // don't use `loose` mode here - need to copy symbols when spreading
+    '@babel/proposal-object-rest-spread',
+    NODE_ENV === 'test' && '@babel/transform-modules-commonjs'
+  ].filter(Boolean)
+}


WayneEllery:
I've re-tested this and it looks like this will fix the issue in edge and IE for `graphql-hooks-memcache` but `tiny-lru` uses classes and `@sindresorhus/fnv1a` uses const which both don't work in IE (but do work in Edge)

bmullan91:
@WayneEllery Could you put the `babelrc.js` in `src` just for consistency, rollup docs recommend that as the location for the config. RE classes, we could include `tiny-lru` to be transpiled by babel; by default, all node_modules are ignored.

WayneEllery:
I'll do that. As for including tiny-lru and @sindresorhus/fnv1a to be transpiled, you would need to change the Rollup config to include the code in the `graphql-hooks-memcache` bundle as currently it's external?

bmullan91:
@wayneclancy Yep - so at the minute this still doesn't work on IE, correct?

We can supply config options to the rollup helper as a second param, I'd treat those as direct overrides to save us doing awkward merging of configs in `~/config/rollup.config.js` - that does mean we'll need to update the memcache config https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks-memcache/rollup.config.js to pass in the correct `external` which will be something like:

```js
import generateRollupConfig from '../../config/rollup.config'

const overrides = {
  external: [
    '@sindresorhus/fnv1a',
    'tiny-lru'
  ]
}

export default generateRollupConfig('GraphQLHooksMemcache', overrides)
```

WayneEllery:
Actually it doesn't work in IE or Edge. This current change (moving the babel config) only fixes the code in `graphql-hooks-memcache`, not the external code in `tiny-lru` and `@sindresorhus/fnv1a`. I'll have a play with the external setting and see what I can do. 

The correct fix would be to fix it in those repos but I think `@sindresorhus/fnv1a` is specifically built for NodeJS and not the browser. For `tiny-lru` it may be possible to fix it there but I'll see.

WayneEllery:
I think it was the spread in object that was causing Edge to break: https://caniuse.com/#feat=mdn-javascript_operators_spread_spread_in_object_literals

bmullan91:
@WayneEllery I have some spare capacity this week to help - do you want me to take this over or would you like to continue working on it?

WayneEllery:
@bmullan91 That would be great. I've been quite busy recently so haven't got round to it

bmullan91:
Closing in favour of #445 - @wayneclancy I'll add you as a co-author when I do the squash & merge.

PR #448, opened 2020-01-22
Make TGraphQLError default to "object"

brookback:
### What does this PR do?

This fixes a breaking change introduced in 4.4.0 in #425.

#425 made the type `UseClientRequestResult` (and other functions) require a second type argument: `TGraphQLError`. To avoid a breaking change in a minor release, it should default to `object` if the user don't want to provide a specific `TGraphQLError` type upfront.

### Checklist


bmullan91:
@brookback Good catch - I've just release this in `graphql-hooks@4.4.1`

brookback:
Cheers for quick release! 👾

PR #465, opened 2020-02-21
Add ability to pass client

### LINKED ISSUE #374: Allow passing client as an option
Similar issue here: https://github.com/nearform/graphql-hooks/issues/315

### Description
As in the linked issue, I need to make requests to different endpoints. The solution in the linked issue was to just add a new HOC that wraps the current component in a `ClientContext.Provider`. While this does work, it more or less goes contrary to the goals of hooks (at this point, you now have a HOC AND hooks, so why not just go with a HOC only?). Also, if you have to make multiple requests in the same component to different endpoints, it's not even possible in this manner.

### Suggested implementation
The easiest (and imo more elegant) solution would be to just allow passing a client as an option.
For example, in useQuery:
```diff
- const client = React.useContext(ClientContext)
+ const client = opts.client || React.useContext(ClientContext)
```
Then, the user can decide how they get their client (e.g. a separate Context).

All in all, it's a very simple change, but allows multiple endpoints rather easily.

jackh726:
### What does this PR do?

Allows passing a `GraphqlClient` to `useQuery`, `useSubscription`, and `useMutation`. It overrides any `ClientContext`.

### Related issues

Fixes #374

### Checklist

### DIFF
--- a/packages/graphql-hooks/index.d.ts
+++ b/packages/graphql-hooks/index.d.ts
@@ -77,7 +77,7 @@ export interface SubscriptionRequest {
 }
 
 export function useSubscription(
-  operation: Operation,
+  operation: UseSubscriptionOperation,
   callback: (response: Result) => void
 ): void
 
@@ -148,6 +148,7 @@ interface UseClientRequestOptions<Variables = object> {
   skipCache?: boolean
   fetchOptionsOverrides?: object
   updateData?(previousData: any, data: any): any
+  client?: GraphQLClient
 }
 
 interface UseQueryOptions<Variables = object>
@@ -170,6 +171,10 @@ interface UseQueryResult<
   refetch(
     options?: UseQueryOptions<Variables>
   ): Promise<UseClientRequestResult<ResponseData, TGraphQLError>>
+}
+
+interface UseSubscriptionOperation extends Operation {
+  client?: GraphqlClient
 }
 
 type FetchData<ResponseData, Variables = object, TGraphQLError = object> = (
--- a/packages/graphql-hooks/src/useClientRequest.js
+++ b/packages/graphql-hooks/src/useClientRequest.js
@@ -72,7 +72,8 @@ function useClientRequest(query, initialOpts = {}) {
     )
   }
 
-  const client = React.useContext(ClientContext)
+  const contextClient = React.useContext(ClientContext)
+  const client = initialOpts.client || contextClient
   const isMounted = React.useRef(true)
   const activeCacheKey = React.useRef(null)
   const operation = {
--- a/packages/graphql-hooks/src/useQuery.js
+++ b/packages/graphql-hooks/src/useQuery.js
@@ -9,7 +9,8 @@ const defaultOpts = {
 
 function useQuery(query, opts = {}) {
   const allOpts = { ...defaultOpts, ...opts }
-  const client = React.useContext(ClientContext)
+  const contextClient = React.useContext(ClientContext)
+  const client = opts.client || contextClient
   const [calledDuringSSR, setCalledDuringSSR] = React.useState(false)
   const [queryReq, state] = useClientRequest(query, allOpts)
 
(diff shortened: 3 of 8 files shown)


packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -72,7 +72,8 @@ function useClientRequest(query, initialOpts = {}) {
     )
   }
 
-  const client = React.useContext(ClientContext)
+  const contextClient = React.useContext(ClientContext)
+  const client = initialOpts.client || contextClient
```
bmullan91: Minor:

```suggestion
  const client = initialOpts.client || React.useContext(ClientContext)
```

packages/graphql-hooks/src/useQuery.js
```diff
@@ -9,7 +9,8 @@ const defaultOpts = {
 
 function useQuery(query, opts = {}) {
   const allOpts = { ...defaultOpts, ...opts }
-  const client = React.useContext(ClientContext)
+  const contextClient = React.useContext(ClientContext)
+  const client = opts.client || contextClient
```
bmullan91: Minor:

```suggestion
  const client = opts.client || React.useContext(ClientContext)
```

packages/graphql-hooks/src/useSubscription.js
```diff
@@ -6,7 +6,8 @@ function useSubscription(options, callback) {
   const callbackRef = useRef(callback)
   callbackRef.current = callback
 
-  const client = useContext(ClientContext)
+  const contextClient = useContext(ClientContext)
+  const client = options.client || contextClient
```
bmullan91: Minor:

```suggestion
  const client = options.client || useContext(ClientContext)
```

packages/graphql-hooks/test/unit/useClientRequest.test.js
```diff
@@ -43,6 +55,21 @@ describe('useClientRequest', () => {
     expect(state).toEqual({ cacheHit: false, loading: true })
   })
 
+  it('uses a passed client', async () => {
+    const options = { isMutation: false, client: mockClient2 }
```
bmullan91: Minor: Could you define the `mockClient2` inside this test.

packages/graphql-hooks/index.d.ts
```diff
@@ -172,6 +173,10 @@ interface UseQueryResult<
   ): Promise<UseClientRequestResult<ResponseData, TGraphQLError>>
 }
 
+interface UseSubscriptionOperation extends Operation {
```
bmullan91: Perhaps we should rename this to `UseSubscriptionOptions`

packages/graphql-hooks/src/useClientRequest.js
jackh726: You can't do this because then you have conditional hooks.

packages/graphql-hooks/index.d.ts
jackh726: Leave `Operation` as is?

jackh726: One point in favor of *not* changing it to `useSubscriptionOptions` is that, unlike `useQuery`, the "options" (variables, operationName, client) are in the same object as `query`. I think, in hindsight, making `useSubscription` be more similar to `useQuery` with a signature of `query,options,callback` would have been better.

jackh726: Another point in favor of *not* changing it (at least not now), is that it's referred to as `Operation` or `operation` in a couple places (like `README`).

packages/graphql-hooks/src/useClientRequest.js
bmullan91: Doh! Completely forgot about that...

packages/graphql-hooks/index.d.ts
bmullan91: @jackh726 Fair points 👏  - Let's leave as is for now.

bmullan91:
@all-contributors please add @brookback for bug code doc test

bmullan91:
@all-contributors please add @jackh726 for bug code doc test

bmullan91:
Available in `graphql-hooks@4.4.3`

PR #493, opened 2020-04-23
feat(persisted): Add support for persisted queries.

salmanm:
### What does this PR do?

- This PR adds a `persisted` option support to `useQuery`. It's useful to tell the backend graphql server if you're sending query SHA instead of the actual query text.

https://github.com/mcollina/fastify-gql/pull/139 adds in support for the same on server side.

### Checklist


packages/babel-plugin-extract-gql/examples/fastify-gql/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "main": "index.js",
+  "scripts": {
+    "start": "node ."
+  },
+  "author": "Salman Mitha <SalmanMitha@gmail.com>",
+  "license": "MIT",
+  "dependencies": {
+    "fastify": "^2.13.1",
+    "fastify-cors": "^3.0.3",
+    "fastify-gql": "salmanm/fastify-gql"
```
mcollina, line 14: This needs to be updated, we merged the code in fastify-gql.

packages/babel-plugin-extract-gql/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "babel-plugin-extract-gql",
+  "version": "2.1.0",
```
mcollina, line 3: Is this correct? How does it work with lerna?

packages/babel-plugin-extract-gql/test/unit/fixtures/actual/.4.js
```diff
@@ -0,0 +1,12 @@
+import { useQuery, persist } from 'graphql-hooks'
```
mcollina, line 1: why is this file `.4.js` instead `4.js`?

.eslintignore
```diff
@@ -3,3 +3,4 @@ dist
 es
 lib
 build
+**/fixtures
```
mcollina, line 6: why are we ignoring them in eslint?

salmanm: Coz the unit test files are partial component code and linter might complain for certain things. Also, `fixtures/expected` contains code that babel outputs, which is not linter compliant.

packages/babel-plugin-extract-gql/test/unit/fixtures/actual/.4.js
salmanm: To disable this test case. Basically this should cover the case where an external options object is supplied to `useQuery` (e.g. `useQuery(quer, optObj)`). The code doesn't yet know how to manipulate that object.

But simpler cases like `useQuery(query)` OR `useQuery(query, { a: 1 })` are covered and working fine.

I was hoping to come back and fix the 4th test.

.eslintignore
mcollina: ok

Issue #494, opened 2020-04-27
Ability to reset the data after useMutation


yanickrochon:
### Description

When `useMutation` returns data, there should be a way to clear that data away as if no calls have been made, that is return the state of the hook to the one it was before invoking the mutation function.

### Suggested implementation

```js
const [ updateFoo, { loading, data, error, reset } = useMutation(FOO_MUTATION);
```

Calling `reset` would clear `data` and `error` and cancel any ongoing request.

mcollina:
would you like to send aPR?

jh3y:
Have you begun working on a solution for this @yanickrochon?
Can we assign this to you? 👍

yanickrochon:
@jh3y  Because of time constraint, I have switched to Apollo server and client for this project. I'm already way behind on schedule and I'm unable to allocate on more at the moment. I have taken the time to look at the code and believe that simply wrapping a function with a reset state (the feature is already there) to revert using the previous context would be a good starting point, but this would need to be tested and this is the part which I have no time for. Not right now anyway.

jh3y:
Completely understandable @yanickrochon 🙂 
Thanks for the prompt response!

PR #503, opened 2020-05-12
feat: adding reset function after useMutation

galmail:
### What does this PR do?

This PR introduces the ability to trigger a reset. See [Issue #494](https://github.com/nearform/graphql-hooks/issues/494)

### Checklist

### DIFF
--- a/packages/graphql-hooks/test/unit/useClientRequest.test.js
+++ b/packages/graphql-hooks/test/unit/useClientRequest.test.js
@@ -68,6 +68,30 @@ describe('useClientRequest', () => {
     )
   })
 
+  it('resets data when reset function is called', async () => {
+    let fetchData, state, resetData
+    renderHook(
+      () => ([fetchData, state, resetData] = useClientRequest(TEST_QUERY)),
+      {
+        wrapper: Wrapper
+      }
+    )
+    // initial state
+    expect(state).toEqual({ cacheHit: false, loading: true })
+    await act(fetchData)
+    expect(state).toEqual({
+      cacheHit: false,
+      loading: false,
+      data: 'data'
+    })
+    await act(resetData)
+    // should be back to initial state
+    expect(state).toEqual({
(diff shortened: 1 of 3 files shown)


packages/graphql-hooks/test/unit/useClientRequest.test.js
```diff
@@ -68,6 +68,30 @@ describe('useClientRequest', () => {
...
+    expect(state).toEqual({ cacheHit: false, loading: true })
+    await act(fetchData)
+    expect(state).toEqual({
+      cacheHit: false,
+      loading: false,
+      data: 'data'
+    })
+    await act(resetData)
+    // should be back to initial state
+    expect(state).toEqual({
```
jh3y, line 89: I'm wondering if there's a case where falling back to the `initialState` completely is not preferable. 
The issue mentions only rolling back the value of `data` in the `state`. Is there a use case where we might only want to roll back the `data` value?

galmail: @jh3y I think it is related to these:
https://github.com/apollographql/apollo-feature-requests/issues/157
https://github.com/apollographql/apollo-feature-requests/issues/170

jh3y: What if we provided a way to pass in a "desired" state that defaults to initial?
That way if a user wants to preserve part of the state, but not all of it, they can.

`resetState({...currentState, data: undefined})`

Or

`resetState()`

galmail: Sounds good, I've now added it, have a look and let me know what you think.

jh3y: Looks good! 👍

jh3y:
Can we add some docs for this though please? 🙏 

And mention the override capability.

PR #478, opened 2020-03-20
Populate error information during SSR

marcelalburg:
The problem is described in #434

### What does this PR do?

During SSR rendering the fetch Promise is stored inside an array and will be resolved later. If an graphql error occurs during this fetch, there is no possibility to see this error during the SSRendering. 

My PR stores the error information in the cache (ONLY FOR SSR) and populate this error information back to the caller.

Issue #434, opened 2019-12-31
Next.js and graphql-hooks with ServerSide Rendering 


marcelalburg:
### Package
`graphql-hooks`
`graphql-hooks-ssr`

### Environment

- `graphql-hooks` version: 4.3.0
- `graphql-hooks-ssr` version: 1.1.5
- `graphql-hooks-memcache` version: 1.3.0
- Browser: SSR

### Description

I use next.js with graphql-hooks and have trouble with the SSR. If you run graphql-hooks on the server and the graphql response contains errors i don't see any possibility to catch this error.

Like i understand graphql-hooks it just store the data inside the cache -> https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks/src/useClientRequest.js#L176

When graphql-hooks-ssr resolves the promises and returns the data from the cache (that not contains any information about the error anymore) -> https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks-ssr/index.js#L23

From my point of view, i cannot render an error page with more information about the graphql error. 

### Suggested solution (optional)

I would like to have a possibility to use the error field on the SSR too or any information how to process the errors on the Server.

Thanks a lot and a happy new year

PR #507, opened 2020-05-19
fix: including error when cached in SSR

galmail:
### What does this PR do?

It is the same idea behind https://github.com/nearform/graphql-hooks/pull/478
The unit test for this scenario is included

### Related issues

Related to [Issue #434](https://github.com/nearform/graphql-hooks/issues/434)

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/useClientRequest.js
+++ b/packages/graphql-hooks/src/useClientRequest.js
@@ -178,6 +178,7 @@ function useClientRequest(query, initialOpts = {}) {
 
           if (client.ssrMode) {
             const cacheValue = {
+              error: actionResult.error,
               data: revisedOpts.updateData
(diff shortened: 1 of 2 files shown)


terryyu:
@galmail is there any plan when it will be released? the issue can cause an infinite loop during SSR when request error happens.
There is a recursive call of [getInitialState](https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks-ssr/index.js#L21), the exit condition is the zero length of `client.ssrPromises`. The guardian condition to populate `client.ssrPromises` is not working in the second call, because the state is reading from cache, and error is not included.
```javascript
// result may already be in the cache from previous SSR iterations
    if (!state.data && !state.error) {
      const p = queryReq()
      client.ssrPromises.push(p)
    }
```

Issue #333, opened 2019-07-17
Refetch Causes Component Reload


Saintroi:
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: 3.6.0
- `graphql-hooks-memcache` version: 1.2.1
- `react` version: 16.8.6
- Browser: Chrome

### Description

I am updating my application to focus on hooks usage, and in the process I wanted to try and replace the GraphQL calls with hooks as well.

I have a component that displays a list of users, and other components with forms that allow creation, editing and removal of those users. I pass the refetch function from the query that gets the list of users down to the other components for them to call after performing a mutation.

With the Apollo tools for querying, this refetch call simply re-populates the list of users without any re-loading of the component or page (no flash or time where there are no users in the list)

However, with this plugin the refetch function seems to re-load the entire component resulting in a momentary blank area where the component should be. This occurs regardless of if I call it from that component or if I pass it to another component to call after a mutation.

### How to reproduce

Refetch a query, watch the component reload.

I figure you guys might be aware of this, maybe it's related to #52 but I wasn't totally sure.

bmullan91:
@Saintroi sorry for the delay; do you might including a code example of this issue you're describing? I think we can solve your issue with the current API :)

Saintroi:
Sure! Code is below, just to confirm I added a "refetch" button on my page that calls the refetch for the LIST_USERS query, which then causes the entire component to reload.

Typically I'd be passing this refetch down to a create/delete/update user component and calling it once those actions have been taken, so want to avoid what is essentially a page refresh.

Thanks!

```js
function ListUser(props) {

  const [searchQuery, setSearchQuery] = useState('');
  const search = React.createRef();

  function handleSearchChange(event) {
    setSearchQuery(event.target.value);
    search.current.focus();
  }

  function handleSearch(users) {
    let newUsers = []
    const str = searchQuery.toLowerCase()

    users.forEach(function(user) {
      if(user.name.toLowerCase().includes(str) || user.cellPhone.toLowerCase().includes(str)) newUsers.push(user);
    })

    return newUsers.sort(compare);
  }

 useEffect(() => {
    console.log("list")
    !dat.loading && search.current.focus();
  });

  let dat = {};

  dat = useQuery(CURRENT_USER);

  !(dat.loading || dat.error || !dat.data.me.admin) && auth.setAdmin();
  !(dat.loading || dat.error) && auth.setId(dat.data.me.id)

[...]

PR #520, opened 2020-06-03
Feature/333 refetch patch option

danwilliams:
### Summary

- Adds a `patch` option to refetch, when set the data / previous result is mutated to match the new result using `deep-diff` (instead of the default behaviour where the new result replaces the previous result)
- This can be used to prevent components from re-rendering (i.e. with `React.memo`) on refetch however will not be as performant as providing an `updateData` handler tailored to the expected data. A note about this has been added to the repo documentation.

### Related issues

https://github.com/nearform/graphql-hooks/issues/333

### Checklist


simoneb:
@danwilliams could you please fix the typo that breaks the build?

simoneb:
Closing as stale

Issue #444, opened 2020-01-21
Data variable from useQuery is briefly undefined – inconsistent with TS types


brookback:
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: 4.3.0
- `react` version: 16.12.0
- Browser: Chrome

### Description

I noticed the `data` variable returned from `useQuery` is undefined when first rendering a component. This collides with the fact that the types say it's always defined.

See `index.d.ts`:

```ts
interface UseClientRequestResult<ResponseData> {
  loading: boolean
  cacheHit: boolean
  data: ResponseData // <- Here
  error?: APIError
}
```

### How to reproduce

```tsx
const { data, loading, error } = useQuery<
        QueryData,
        QueryVariables
    >(QUERY, {
        variables: {},
    });

console.log({ data, loading, error });
```

**Result**

<img width="484" alt="image" src="https://user-images.githubusercontent.com/30257156/72808371-b5519500-3c59-11ea-9fb0-0dd44e0ede18.png">

### Suggested solution

Make `data` optional/falsy on `UseClientRequestResult` if we can't guarantee it being truthy while loading:

```diff
interface UseClientRequestResult<ResponseData> {
  loading: boolean
  cacheHit: boolean
- data: ResponseData
+ data?: ResponseData
  error?: APIError
}
```

brookback:
Note that fixing this is a breaking change, since if people rely on `data` always being defined, making it optional will cause a bunch of compile errors.

jh3y:
Where should we go with this one?

Seems like the `Suggested solution` is one we should merge in 👍 

However, we don't have a `CHANGELOG`? or release notes? To communicate a breaking change like this.

brookback:
@jh3y There is a changelog in https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks/CHANGELOG.md.

However, if this repo is following semantic versioning (which I think it is), the major version must be bumped from `5.x.x`.

jh3y:
Sorry, my bad, I did see that after writing this 🤦 
I'd initially looked at release notes and must've missed it 😅 

Yeah, a major bump sounds like the right way forward here 👍 Thanks!

Issue #517, opened 2020-06-01
Data in useQuery is undefined on first render


vladshcherbin:
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: 4.4.4
- `react` version: 16.13.1
- Browser: chrome

### Description

Data is `undefined` until it's fetched. Docs say it should be an object. While it's `undefined` it's not possible to use this:

```js
const { loading, error, data: { users } } = useQuery(usersQuery)
```

Using this syntax is convenient when you need to later use variable (users) multiple times - check length, map over it, etc.

### How to reproduce

I believe there is no need to reproduce.

### Suggested solution (optional)

Maybe it should be an object always like it's written in the docs? It'll also make an example above work 🤔 

Related typings issue - https://github.com/nearform/graphql-hooks/issues/444

ovhemert:
The `data` can be `undefined` on first render. This has been adjusted in the types. See #567 
Returning it as an object would cause side effects.

PR #521, opened 2020-06-03
Fixed useQuery returning undefined on initial render

danwilliams:
### Summary

- Changes `useQuery` so that if `data` is `undefined` an empty object is returned instead, this usually happens on the initial component render and should also resolve issue #444
- Note that the change might be breaking for some users who are relying on the old behaviour where the data is initially undefined. For example, loose assertions like this will no longer work:

```jsx
if (!data) return <div>Loading</div>
```
### Related issues

https://github.com/nearform/graphql-hooks/issues/517
https://github.com/nearform/graphql-hooks/issues/444

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/useQuery.js
+++ b/packages/graphql-hooks/src/useQuery.js
@@ -34,6 +34,7 @@ function useQuery(query, opts = {}) {
 
   return {
     ...state,
+    data: state.data ? state.data : {},
     refetch: React.useCallback(
       (options = {}) =>
         queryReq({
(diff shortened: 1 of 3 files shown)


lynxtaa:
I'm not sure about this.

1. Returning an empty object will  cause rerender loop if I'm using `data` as a hook dependency.
2. It'll make checks for `data` uglier. Right now, i'm simply using `if (!data) return <Spinner />` 
3. Also uglier type guards:
```jsx
// Before
// data: ResponseData | undefined
return data ? data.users.map(...) : <Spinner />

// After 
// data: ResponseData | {}
return 'users' in data ? data.users : <Spinner />
```

ovhemert:
The issue was fixed in PR #567 to allow `data` be `undefined` in the types.
Returning it as an object would cause side effects.

PR #559, opened 2020-09-14
Add mock GraphQL server

ovhemert:
### What does this PR do?

Adds a mock graphql server to replace graph.cool (since that service is deprecated).
Server is needed to run the `create-react-app` example, but also to run acceptance test.


mcollina (approved):
lgtm

PR #560, opened 2020-09-15
Replace mock server calls

ovhemert:
### What does this PR do?

Fixes the urls to the mock graphql server (previously graphcool, which is deprecated).
And makes acceptance test work again


mcollina (approved):
lgtm

PR #567, opened 2020-09-23
fix: response data is optional

ovhemert:
### What does this PR do?

The `useQuery` is returning `undefined` for `data` on the first render, but this is not reflected in the typescript types. This PR changes the type to optional.

Although there is a suggestion #517 (and a PR #521) to change the `data` to always be an empty object, this would not be preferred, because that would cause a re-render.

### Related issues

- #517 
- #444


mcollina (approved):
lgtm

PR #597, opened 2020-11-03
chore: reduce renovate noise

simoneb:
### What does this PR do?

Reduces Renovate noise by:

- automerging all dependency types
- avoid creating PRs for the bumps which pass CI checks so that no notifications are sent
- label Renovate PRs with a "renovate" label

It also unpins dependencies where they were pinned.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,18 +1,11 @@
 {
-  "extends": [
-    "config:base"
-  ],
-  "rangeStrategy": "replace",
-  "ignorePaths": [],
-  "includePaths": [
-    "package.json",
-    "packages/**",
-    "examples/**"
-  ],
+  "extends": ["config:base"],
+  "includePaths": ["package.json", "packages/**", "examples/**"],
+  "label": "renovate",
   "packageRules": [
     {
-      "updateTypes": ["minor", "patch"],
-      "automerge": true
+      "automerge": true,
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -1,18 +1,11 @@
...
-  ],
+  "extends": ["config:base"],
+  "includePaths": ["package.json", "packages/**", "examples/**"],
+  "label": "renovate",
   "packageRules": [
     {
-      "updateTypes": ["minor", "patch"],
-      "automerge": true
+      "automerge": true,
+      "automergeType": "branch"
```
simoneb, line 8: this change prevents PRs from being created in case CI passes. If CI does NOT pass, a PR is created so devs can look into why CI failed.

renovate.json
```diff
@@ -1,18 +1,11 @@
...
-    "package.json",
-    "packages/**",
-    "examples/**"
-  ],
+  "extends": ["config:base"],
+  "includePaths": ["package.json", "packages/**", "examples/**"],
+  "label": "renovate",
   "packageRules": [
     {
-      "updateTypes": ["minor", "patch"],
```
simoneb, line 14: as suggested by @mcollina , all dependency types are auto merged, including majors.

mcollina (approved):
lgtm

Issue #586, opened 2020-10-23
Move the CI pipeline to Github Actions


mcollina:
As titled.

PR #606, opened 2020-11-30
ci: github actions

simoneb:
### What does this PR do?

Fixes https://github.com/nearform/graphql-hooks/issues/586

Moves CI from Circle CI to GH Actions.

Also moves the acceptance-tests run which is triggered by a successful deploy to Netlify to GH actions, meaning that:

- it creates a GH Action Workflow `acceptance-tests` which is not run automatically but should be triggered manually
- it changes the Netlify function which triggers the acceptance tests to invoke the GitHub API instead of the CircleCI API

### Things to consider

- GH works differently from Circle CI and even though the acceptance tests are run on the right branch the check does not appear as a "check" inside the PR that triggered id. It does appear in the Actions tab though, but it will not prevent a PR from being merged. I don't think there's a way around this
- The checks are red because I won't remove the Circle CI check until this PR is merged. Other checks should be green though


.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+    env:
+      COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache node modules
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: npm-${{ hashFiles('**/package.json') }}
```
codyzu, line 26: I realize the root package does not have one, but would it be better to also include the `package-lock.json` files? That way the cache invalidates on changes to transient dependencies, not just the deps in the `package.json`. :thinking:

simoneb: I agree, but since I don't know why the root package does not have one and since it's [intentionally excluded](https://github.com/nearform/graphql-hooks/blob/master/.npmrc), I didn't want to modify it as the reason why it's excluded may be valid but we may not know it.

simoneb: Funny though, because tracing back to that change where lockfile was removed, it seems there was an overall intention to remove all package lock files, even from packages. But packages do have them now, so apparently that intention was lost over time. That's one reason why I'm cautious reverting existing decisions because we may not know why they were made.

codyzu: Strange.

In the end, it's probably not wise to make a change as I suggested in this PR. Probably best to stick with the original functionality, just as you've done.

simoneb: @codyzu can you approve the PR then please?

codyzu (approved):
LGTM

PR #609, opened 2020-12-16
fix: remove Buffer support from fnv1a

### LINKED ISSUE #608: [memcache] Buffer is not defined (in browser)
### Package

`graphql-hooks-memcache`

### Environment

- `graphql-hooks` version: 5.0.0
- `graphql-hooks-memcache` 2.0.0
- `react`: 16.13.1
- `Browser`: Chrome 88

### Description

```
ReferenceError: Buffer is not defined
fnv1a
.my-project/node_modules/graphql-hooks-memcache/es/graphql-hooks-memcache.js:101
   98 | }
   99 | 
  100 | function fnv1a(input) {
> 101 |   if (input instanceof Buffer) {
  102 |     return fnv1aBuffer(input);
  103 |   } else if (typeof input === 'string') {
  104 |     return fnv1aString(input);
```

### How to reproduce

Simply having the memcache library in the page causes the error. I didn't see any note that the library is Node only, is it?

```jsx
import { GraphQLClient } from "graphql-hooks";
import memCache from "graphql-hooks-memcache";

  url: "/graphql",
  cache: memCache(),
});
```

simoneb:
### What does this PR do?

Removes support for Buffer in the imported fnv1a library as we don't use it and because it causes issues when Buffer is not polyfilled, which happens by default in Webpack 5.

### Related issues

Fixes #608 

### Checklist

### DIFF
--- a/packages/graphql-hooks-memcache/src/fnv1a.js
+++ b/packages/graphql-hooks-memcache/src/fnv1a.js
@@ -37,26 +37,10 @@ function fnv1aString(string) {
   return hash >>> 0
 }
 
-function fnv1aBuffer(buf) {
-  let hash = OFFSET_BASIS_32
-
-  for (let i = 0; i < buf.length; ) {
-    hash ^= buf[i++]
-
-    // 32-bit FNV prime: 2**24 + 2**8 + 0x93 = 16777619
-    // Using bitshift for accuracy and performance. Numbers in JS suck.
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

stefanoruth:
Can we please have a release of this package so we can use this without any hacks for webpack 5 😄

Issue #613, opened 2021-01-05
Migrate builds to Node LTS


simoneb:
Make sure to update the branch protection rules to request the new build instead of the old one as the requirement for merging PRs. If you don't have privileges to do so, ask an owner of the repo before merging your PR

PR #619, opened 2021-02-10
chore: use node 14 LTS

Peterabsolon:
### What does this PR do?

Upgrades Node 12 to 14.

### Related issues

Closes Closes https://github.com/nearform/graphql-hooks/issues/613

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x]
+        node-version: [14.x]
 
     steps:
       - run: echo Running on target url $ACCEPTANCE_URL
--- a/examples/create-react-app/Dockerfile
+++ b/examples/create-react-app/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12-alpine
+FROM node:14-alpine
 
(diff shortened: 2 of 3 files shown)


.github/workflows/acceptance-tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x]
+        node-version: [14.x]
```
Peterabsolon, line 19: @simoneb You said "instead" in the issue so I replaced it, instead of adding it. Is that ok?

examples/create-react-app/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:12-alpine
+FROM node:14-alpine
```
Peterabsolon, line 1: Updated it here too, tested it and it still works, unsurprisingly.

mcollina (approved):
lgtm

.github/workflows/acceptance-tests.yml
simoneb: Yes, because these are client packages so we don't need to make sure their build runs on multiple versions of Node, as Node is not their runtime

simoneb (approved):

PR #622, opened 2021-02-16
Don't require fetch polyfill if ssrMode is false

### LINKED ISSUE #620: Fetch polyfill required even when ssr is configured false
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: `4.0.0`

### Description

Currently if the Client is created on the server without a fetch polyfill it results in error. This happens even when `config.ssr` is `false`. This is an issue with universal code as it means that if the same code that runs on the server it will fail even if a graphql request isn't made.

```
if (!config.fetch && !fetch) {
                          ^

ReferenceError: fetch is not defined
```

### How to reproduce

Create Client in NodeJS:

```
  url: 'https://test/graphql',
  cache: memCache(),
});
```

Only require fetch to be available if `ssr` is enable in the `config` for the client

penx:
### What does this PR do?

Don't throw an error if fetch is not polyfilled and ssrMode is set to false.

### Related issues

Fixes #620

### Checklist

- [ ] I have added or updated any relevant documentation N/A

### DIFF
--- a/packages/graphql-hooks/src/GraphQLClient.js
+++ b/packages/graphql-hooks/src/GraphQLClient.js
@@ -1,9 +1,7 @@
 import { extractFiles } from 'extract-files'
 
 import isExtractableFileEnhanced from './isExtractableFileEnhanced'
-
-const FETCH_POLYFILL_ERROR =
-  'GraphQLClient: fetch must be polyfilled or passed in new GraphQLClient({ fetch })'
+import canUseDOM from './canUseDOM'
 
 class GraphQLClient {
   constructor(config = {}) {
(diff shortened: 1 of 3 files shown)


packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -182,6 +184,9 @@ class GraphQLClient {
   }
 
   request(operation, options = {}) {
+    if (!this.fetch) {
+      throw new Error(FETCH_POLYFILL_ERROR)
```
penx, line 188: I plan to add a test for this error being thrown

simoneb (commented):
The code overall LGTM, but I can't get the app to work on my android device. Clicking the login button shows a warning at the bottom with an error that's not very telling and then does nothing.

simoneb (approved):
LGTM

PR #649, opened 2021-06-11
feat: support `graphql-ws`

### LINKED ISSUE #605: Usage with graphql-ws
### Description

Would be great if the library could be used with [`graphql-ws`](https://github.com/enisdenjo/graphql-ws), which is GraphQL over WebSocket implementation.

### Suggested implementation

Current `GraphQLClient` implementation uses fetch, which requires `url` param to be HTTP(S) requests. A possible implementation is the ability to use ws(s) protocol here as well.

enisdenjo:
### What does this PR do?

Adds support for `graphql-ws`. The changes are actually minimal.

Additionally, I left a [TODO about not handling subscription errors](https://github.com/enisdenjo/graphql-hooks/blob/e6587c44f40913f682f71bce998bdcd2a4e4a1a1/packages/graphql-hooks/src/useSubscription.js#L25). Would like to know more about it, why are they being ignored?

### Related issues

Closes #605 

### Checklist

- [ ] ~I have added or updated any relevant tests~


simoneb (changes requested):
Please 

- use npm@6 to install packages and avoid the lock file issue
- add tests

enisdenjo:
I've added graphql-ws with npm@6.

What exactly do you want to test? There is no unit test for the existing `subscriptions-transport-ws` either, you just mock a client and test the `useSubscription` hook itself.

simoneb:
> I've added graphql-ws with npm@6.
> 
> What exactly do you want to test? There is no unit test for the existing `subscriptions-transport-ws` either, you just mock a client and test the `useSubscription` hook itself.

Existing tests for the `useSubscription` hook are [here](https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks/test/unit/useSubscription.test.js) and run against a mocked implementation of the client which presumably mimics the behavior of subscriptions-transport-ws. Would it not make sense to have the same tests running against something which behaves like graphql-ws instead?

enisdenjo:
Yeah, I've seen the mocked implementation, which is why I am asking. If you noticed how I added `graphql-ws` support, [I literally adapted the lib's API to the `GraphQLClient.createSubscription` (mock).](https://github.com/enisdenjo/graphql-hooks/blob/c71990d78932b706f40c87fb39ec45d8a299614d/packages/graphql-hooks/src/GraphQLClient.js#L256-L266) 

Everything else is pretty much untouched and unchanged. In essence, I dont think there is anything to test really... If you, however, add an integration test in the future - testing the implementation of `graphql-ws` in addition to `subscriptions-transport-ws` would then make sense.

simoneb (approved):

enisdenjo:
The client side is pretty much the same. I've updated the README file to include the new supported client of `graphql-ws`. 

However, I must say that `graphql-ws` protocol (you can check it out [here](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md)) is different from `subscriptions-transport-ws`. So, the example you have which uses `fastify-gql` (now `mercurius`) on the server side **wont** work with `graphql-ws`. But, since this lies namely on the backend, the `graphql-hooks` side stays exactly the same.

simoneb:
> The client side is pretty much the same. I've updated the README file to include the new supported client of `graphql-ws`.
> 
> However, I must say that `graphql-ws` protocol (you can check it out [here](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md)) is different from `subscriptions-transport-ws`. So, the example you have which uses `fastify-gql` (now `mercurius`) on the server side **wont** work with `graphql-ws`. But, since this lies namely on the backend, the `graphql-hooks` side stays exactly the same.

Fair enough. What do you think about having an example for that too, since it's actually quite a few changes?

enisdenjo:
Sure. But, TBH, I dont really have the time for it... So maybe a task for a separate PR or someone else? 😬

simoneb:
Fair enough, it would be a nice to have but this contribution is already good as-is, thanks for this.

PR #663, opened 2021-07-30
feat: add ability to skip query entirely

### LINKED ISSUE #154: Add ability to skip query entirely
### Description

In some cases one could skip making the request to graphql if some condition is true. For example if you have two cascade dropdowns on initial render you want to load and display data only in the first one and the second one just display it as disabled. When you pick something in the first dropdown you have to query the data for the second dropdown.

In this example the second dropdown will skip making the request to get the data until the condition to have selected some value in the first dropdown is met.

### Suggested implementation

A `skip` boolean prop could be set as part of `useQuery` `opts` and if true the entire query will be skipped and it will return `data` as undefined

clemarc:
### What does this PR do?

- Adds a  `skip` option to `useQuery` to offer the ability to skip a query conditionally
- The `skip` option is also watched for changes so the behavior of `useQuery` adapts to changing values of `skip`
- It also works with SSR
- It doesn't affect or change the `refetch` behavior in any way

### Related issues

Closes #154 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ This is a custom hook that takes care of fetching your query and storing the res
   - `operationName`: If your query has multiple operations, pass the name of the operation you wish to execute.
   - `persisted`: Boolean - defaults to `false`; Pass `true` if your graphql server supports `persisted` flag to serve persisted queries.
   - `useCache`: Boolean - defaults to `true`; cache the query result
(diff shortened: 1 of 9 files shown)


mcollina (approved):
lgtm

clemarc (commented):
Thanks @simoneb for your help :)

PR #665, opened 2021-08-03
feat: throwErrors option in useQuery

### LINKED ISSUE #324: throw errors from hooks
same reasoning as in https://github.com/trojanowski/react-apollo-hooks/issues/28

simoneb:
### What does this PR do?

Adds a `throwErrors` option to `useQuery` which throws errors instead of returning an `error` property in the resulting object.

This allows using Error Boundaries to intercept and display errors.

### Related issues

Resolves #324 

### How to test

See the updated create-react-app example application

### Checklist

### DIFF
--- a/examples/create-react-app/src/App.js
+++ b/examples/create-react-app/src/App.js
@@ -1,7 +1,9 @@
 import React, { useState } from 'react'
 import { GraphQLClient, ClientContext } from 'graphql-hooks'
 import memCache from 'graphql-hooks-memcache'
+
 import Posts from './components/Posts'
+import PostsWithErrorBoundary from './components/PostsWithErrorBoundary'
 
 const client = new GraphQLClient({
   cache: memCache(),
@@ -19,6 +21,7 @@ export default function App() {
       <span>Verify caching by hiding/showing the posts: </span>
       <button onClick={togglePosts}>{showPosts ? 'hide' : 'show'}</button>
       {showPosts && <Posts />}
+      {showPosts && <PostsWithErrorBoundary />}
     </ClientContext.Provider>
   )
 }
--- a/examples/create-react-app/src/components/CreatePost.js
+++ b/examples/create-react-app/src/components/CreatePost.js
@@ -1,5 +1,5 @@
 import { useMutation } from 'graphql-hooks'
-import PropTypes from 'prop-types'
+import T from 'prop-types'
 import React from 'react'
 import CreatePostForm from './CreatePostForm'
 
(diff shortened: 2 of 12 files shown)


machineghost:
It looks pretty good to me (didn't do a super thorough review but I looked at the entire commit).  My only meaningful thought was that:

      if (!data || !data.allPosts || !data.allPosts.length) return 'No posts'

could be simplified using modern JS:

      if (!data?.allPosts?.length) return 'No posts'

I'd also suggest (and this is pure opinion here) that I consider PropTypes an anti-pattern in 2021.  If you want typing, PropTypes are awful: use TypeScript instead.  Conversely, if you want to us a dynamically typed language like JS, shoehorning-in typing for only a small fraction of the variables in your app seems like a waste of time.

But again, if you <3 PropTypes (and it seems you do), don't let me stop you: the commit looks great.

simoneb:
> It looks pretty good to me (didn't do a super thorough review but I looked at the entire commit). My only meaningful thought was that:
> 
> ```
>   if (!data || !data.allPosts || !data.allPosts.length) return 'No posts'
> ```
> 
> could be simplified using modern JS:
> 
> ```
>   if (!data?.allPosts?.length) return 'No posts'
> ```
> 
> I'd also suggest (and this is pure opinion here) that I consider PropTypes an anti-pattern in 2021. If you want strong typing, PropTypes are an awful way to get it: use TypeScript instead. Conversely, if you want to us a dynamically typed language like JS, shoehorning-in typing for only a small fraction of the variables in your app seems like a waste of time.
> 
> But again, if you <3 PropTypes (and it seems you do), don't let me stop you: the commit looks great :)

Let's not focus on this right now, all that code comes from simply copying existing code, I'm more interested in feedback about the actual feature.

machineghost (commented):
As I said in my comment (which I accidentally made as a comment, and then deleted since I meant to make it a review comment), all I had were two subjective points (using modern JS could improve that one line, and using PropTypes in 2021 is arguably an anti-pattern).

As far as the fundamental code of the feature, it looked great to me!  I'm not at all a major maintainer on this project though, just some guy that submitted a commit once (and is now just offering "one more set of eyes"), so please take this with the requisite grain of salt and wait for a "real" maintainer (thus my only adding this as a review comment and not an approval).

simoneb:
I appreciate the feeback!

> wait for a "real" maintainer

I'm one, but that doesn't mean I merge code without asking feedback from those who suggested the feature in the first place :)

machineghost:
>I'm one, but that doesn't mean I merge code without asking feedback from those who suggested the feature in the first place :)

Ah; I wasn't sure (thought you might just be a random contributor like myself).  In any case, you got my two cents :)

Issue #682, opened 2021-11-08
Errors are silently ignored in `useSubscription`


bitjson:
### Package

`graphql-hooks`

### Description

The `useSubscription` hook silently ignores any received errors:

https://github.com/nearform/graphql-hooks/blob/b7ff57d30fa9894162e7145bc5a5ad98e0561f0b/packages/graphql-hooks/src/useSubscription.js#L20-L31

It looks like this was discussed in https://github.com/nearform/graphql-hooks/pull/649: @simoneb or @enisdenjo, did either of you have a preference for how error handling should work? The readme documents an `errors` property in the callback, is that still the preferred API?

https://github.com/nearform/graphql-hooks/blob/b7ff57d30fa9894162e7145bc5a5ad98e0561f0b/README.md#L427-L432

@enisdenjo I see you're [improving the documentation around error handling in `graphql-ws`](https://github.com/enisdenjo/graphql-ws/commit/216075510ad7ecbe2fdd4ba6fe578a668ae38fe3), but I'm not sure if I fully understand – should errors always be a JS `Error` object? Or can `graphql-ws` sometimes return an array of errors? 

Thanks!

bitjson:
Ok, I was unsure about error handling because of [a bug in Hasura's implementation of the `graphql-ws` protocol](https://github.com/hasura/graphql-engine/issues/7786). With that tracked down, I think the error handling strategy from the readme is still probably the right API, so I implemented it and fixed the tests in PR https://github.com/nearform/graphql-hooks/pull/683. Reviews appreciated!

enisdenjo:
> It looks like this was discussed in #649: @simoneb or @enisdenjo, did either of you have a preference for how error handling should work? The readme documents an errors property in the callback, is that still the preferred API?

I am not an user of `graphql-hooks` and therefore have no preference. I left the TODO comment because errors are **important**, the lib's maintainers should take them into count.

> Should errors always be a JS Error object? Or can graphql-ws sometimes return an array of errors?

The error can be literally _anything_. I pushed https://github.com/enisdenjo/graphql-ws/commit/216075510ad7ecbe2fdd4ba6fe578a668ae38fe3 because the conditionals ordering was wrong and that lead to matching problems - if you take a closer look at the commit you'll notice that only the order changed. Read more about error handling here: https://github.com/enisdenjo/graphql-ws/discussions/226#discussioncomment-1216935.

PR #683, opened 2021-11-08
fix: handle errors in useSubscription

bitjson:
### What does this PR do?

Instead of silently ignoring errors sent from the server, this PR makes `useSubscription` handle errors in the way documented (but previously unimplemented) in the readme:

https://github.com/nearform/graphql-hooks/blob/b7ff57d30fa9894162e7145bc5a5ad98e0561f0b/README.md#L427-L432

### Related issues

Fixes https://github.com/nearform/graphql-hooks/issues/682

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/useSubscription.js
+++ b/packages/graphql-hooks/src/useSubscription.js
@@ -21,9 +21,8 @@ function useSubscription(options, callback) {
       next: result => {
         callbackRef.current(result)
       },
-      error: () => {
-        // TODO-db-210611 errors are important, why not handle them?
-        subscription.unsubscribe()
+      error: errors => {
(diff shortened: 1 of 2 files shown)


simoneb (approved):
This change LGTM, I guess it's going to be a breaking change though, since previously errors would lead to a subscription being unsubscribed.

What do you think will be the impact on this change on existing users?

enisdenjo:
I dont think this would be a breaking change. Simply because receiving an `Error` message is as terminating as receiving a `Complete` message in `graphql-ws`. It tends to mirror the observable pattern. In both cases, the subscription gets unsubscribed.

enisdenjo (approved):

Issue #74, opened 2019-02-28
Refetch queries with mutations subscription


jgoux:
Hello,

### Description

One important feature of a graphql-client to me is the ability to refetch/update the data of a query after one or multiple mutations executions.

I really liked the idea of "a query subscribes to one or more mutations" here (see onMutation parameter) : https://github.com/arackaf/micro-graphql-react#building-queries

I started playing with the idea that we could leverage the `useEffect` wrapping the query fetching call and pass extra inputs to its second argument to automatically refetch the query again.

The extra inputs would take the form of a list of mutations results so whenever a subcribed mutation happens, the extra inputs changed and the query is fired again.

### Suggested implementation
```
useQuery(GET_TODO, { id }, { refetch: ['addTodo', 'updateTodo', 'deleteTodo'] })
```

I made a working example of the idea here : https://codesandbox.io/s/m3nq0315m9

The general idea is to maintain of global registry of mutations results in the Context.
From here, `useQuery` could accept a `refetch` options taking a list of mutations names. Then it's just a matter of mapping the list of mutations names into a list of mutations results, and voilà, we have our extra inputs to give to `useEffect`.

The approach is naive but quite effective, we could complexify it a bit by accepting a tuple with an extra function given the mutation result, so we have control over the generated input given to useEffect :
```
useQuery(GET_TODO, { id }, { refetch: [['removeTodo', ({ id: removedId }) => id === removedId]] })
```

Joezo:
We've been thinking about different ways of solving this problem. I'm not against the way you've suggested although I think I would prefer telling the mutations which query they should update, rather than this way around.

There's a related issue here #52.

This is very much open for discussion at the moment.

jgoux:
@Joezo Could you elaborate about why you prefer to do the subscriptions on the mutation side?

As I see it, the mutation is a data producer, and the queries are data consumers. Putting these dependencies on the data producer feels weird as it will have to know about every consumers. On the contrary, making each consumer decide for itself when it has to be refetch seems more flexible.
Generally there are far more read operations than write in an application. I prefer editing a few queries if I add a mutation than the other way around.

bmullan91:
@jgoux This is a very interesting idea, I've always thought of the dependency as mutation 1 - many queries though.

I could see this working well for simple queries/mutation with no arguments - like the todo's example you gave. But what about mutations and queries with arguments how would to correctly verify that a query should be refetched?

[...]

Issue #52, opened 2019-02-25
Add ability to update the cache after a mutation


jackdclark:
Currently the only way to update a record after a mutation is to refetch the entire query.

bmullan91:
I've been thinking about this + auto refetching after a mutation. Here are the three methods that's been suggested so far:

1. `useMutation(MUTATION, { updateCache })`
2. `useMutation(MUTATION, { refetchQueries })`
3. `useQuery(QUERY, { refetchAfterMutations })`

---

# updateCache

`useMutation(MUTATION, { updateCache })`

## Usage

```js
import { LIST_USERS_QUERY } from "./somewhere";

const CREATE_USER_MUTATION = `...`;

function CreateNewUser() {
  const [createUser] = useMutation(CREATE_USER_MUTATION, {
    updateCache: (cache, result) => {
      const cacheKey = client.getCacheKey({ query: LIST_USERS_QUERY });
      const existingResult = cache.get(cacheKey);
      cache.set(cacheKey, {
        ...existingResult,
        data: {
          users: [...existingResult.data.users, result.data.createUser]
        }
      });
    }
  });
}
```

## Implementation

```js
// sumarised useClientRequest.js
fetchData() {
  return this.client.request(operation, options).then(result => {
    if (options.updateCache) {
      options.updateCache(cache, result)
    }
  })
}
```

For this to work correctly we need a way to re-render components that would be affected by the cache change.

Adding an event emitter interface to the `cache` would enable this. We would emit the following events:

- `cache.set(key, data) -> emitter.emit(key, data)`
- `cache.delete(key) -> emitter.emit(key)`
- `cache.clear() -> emitter.emit('CACHE_CLEARED')`

For the component to re-render, we need an additional `useEffect` inside `useQuery` that registers the event handlers:

```js
// summarised useQuery.js
const [queryReq, state] = useClientRequest(query, allOpts);

React.useEffect(() => {
  if (!client.cache || !client.cache.on) {
    return;
  }

  // This implies we'd also store the cacheKey in the
  // useClientRequest state
  client.cache.on(state.cacheKey, queryReq);
  client.cache.on("CACHE_CLEARED", queryReq);
  return () => client.cache.off(state.cacheKey, queryReq);
}, [state.cacheKey]);
```

---

# refetchQueries

`useMutation(MUTATION, { refetchQueries })`

This is a feature that developers are familiar with in Apollo:

[...]

PR #686, opened 2021-11-17
feat: refetch query when subscribed mutation executes

giacomorebonato:
### Refetch a query when a dependant mutation executes

I read the following issues:

> "Add ability to update the cache after a mutation #52"

> Refetch queries with mutations subscription #74

They contain a proposal by @bmullan91 and jgoux that I liked, the n.3 about **refetchAfterMutations**.

Meaning that when using `useQuery` we can subscribe to a list of mutations that make the entire query refetch.
It also allows a `filter` property with a function that allows refetch only when it evaluates true.

Query below executes again when `UPDATE_USER_MUTATION` executes and contains `userId === 1` in its variables.

```js
useQuery(TEST_QUERY, {
  refetchAfterMutations: [
    {
      mutation: UPDATE_USER_MUTATION,
      filter: ({ userId }) => userId === 1
    }
  ]
})
```

The above is achieved by adding an EventEmitter ([new events dependency](https://www.npmjs.com/package/events)) property to the GraphQLClient class.

While the issue also mentions the manual manipulation of the cache when a mutation happens to avoid roundtrips, I believe that this implementation stays relevant because it's easier to use and flexible.

It's my first contribution to the project and I am afraid that I am missing out something important: looking forward to hearing your thoughts :)

### Related issues

- https://github.com/nearform/graphql-hooks/issues/52
- https://github.com/nearform/graphql-hooks/issues/74

### For reviewers

The first commit in this work contains the implementation which touches 8 files.  
The next commits is me updating the project to Node 16, to fix Netlify build that relies already on it.
If this work wasn't needed I'll take them out.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,9 @@ This is a custom hook that takes care of fetching your query and storing the res
     - `previousData`: Previous GraphQL query or `updateData` result
     - `data`: New GraphQL query result
   - `client`: GraphQLClient - If a GraphQLClient is explicitly passed as an option, then it will be used instead of the client from the `ClientContext`.
+  - `refetchAfterMutations`: Array of RefetchAferMutationsData - Mutations that will cause the refetch of the query
+    - `mutation`: String - The mutation string
+    - `filter`: Function (optional) - It receives mutation variables as parameter and blocks refetch if it returns false
 
 ### `useQuery` return value
 
@@ -573,6 +576,50 @@ export default function PostList() {
   )
 }
 ```
+
+## Refetch queries with mutations subscription
+
+We can have a query to automatically refetch when any mutation from a provided list execute.  
+In the following example we are refetching a list of posts for a given user.
+
+**Example**
+
+```jsx
+export const allPostsByUserIdQuery = `
+  query allPosts($userId: Int!) {
+    allPosts(userId: $userId) {
+      id
+      title
+      url
+    }
+  }
+`
+
+export const createPostMutation = `
+  mutation createPost($userId: Int!, $text: String!) {
+    createPost(userId: $userId, text: $text) {
+      id
+      title
+      url
+    }
+  }
+`
+
+const myUserId = 5
+
+useQuery(allPostsByUserIdQuery, {
+  variables: {
+    userId: myUserId
+  },
+  refetchAfterMutations: [
+    {
+      mutation: createPostMutation,
+      filter: (variables) => variables.userId === myUserId
+    }
+  ]
+})
+```
+
 
 ## File uploads
 
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ This is a custom hook that takes care of fetching your query and storing the res
     - `data`: New GraphQL query result
   - `client`: GraphQLClient - If a GraphQLClient is explicitly passed as an option, then it will be used instead of the client from the `ClientContext`.
   - `refetchAfterMutations`: list of mutations that will trigger the query refetch. It can be either:
-    - String | String[] - The string of the mutation or an array of them
+    - String | String[] - The string containing the mutation's definition or an array of them
     - Object | Object[] with the following properties or an array of them
       - `mutation`: String - The mutation string
       - `filter`: Function (optional) - It receives mutation variables as parameter and blocks refetch if it returns false
--- a/README.md
+++ b/README.md
@@ -243,9 +243,11 @@ This is a custom hook that takes care of fetching your query and storing the res
     - `previousData`: Previous GraphQL query or `updateData` result
     - `data`: New GraphQL query result
   - `client`: GraphQLClient - If a GraphQLClient is explicitly passed as an option, then it will be used instead of the client from the `ClientContext`.
-  - `refetchAfterMutations`: Array of RefetchAferMutationsData - Mutations that will cause the refetch of the query
-    - `mutation`: String - The mutation string
-    - `filter`: Function (optional) - It receives mutation variables as parameter and blocks refetch if it returns false
+  - `refetchAfterMutations`: list of mutations that will trigger the query refetch. It can be either:
+    - String | String[] - The string of the mutation or an array of them
+    - Object | Object[] with the following properties or an array of them
+      - `mutation`: String - The mutation string
+      - `filter`: Function (optional) - It receives mutation variables as parameter and blocks refetch if it returns false
 
 ### `useQuery` return value
 
--- a/README.md
+++ b/README.md
@@ -243,11 +243,12 @@ This is a custom hook that takes care of fetching your query and storing the res
     - `previousData`: Previous GraphQL query or `updateData` result
     - `data`: New GraphQL query result
   - `client`: GraphQLClient - If a GraphQLClient is explicitly passed as an option, then it will be used instead of the client from the `ClientContext`.
-  - `refetchAfterMutations`: list of mutations that will trigger the query refetch. It can be either:
-    - String | String[] - The string containing the mutation's definition or an array of them
-    - Object | Object[] with the following properties or an array of them
+  - `refetchAfterMutations`: String | Object | (String | Object)[] - You can specify when a mutation should trigger query refetch.
+    - If it's a string, it's the mutation string
+    - If it's an object then it has properties mutation and filter
       - `mutation`: String - The mutation string
-      - `filter`: Function (optional) - It receives mutation variables as parameter and blocks refetch if it returns false
+      - `filter`: Function (optional) - It receives mutation's variables as parameter and blocks refetch if it returns false
+    - If it's an array, the elements can be of either type above
 
 ### `useQuery` return value
 
--- a/examples/create-react-app/src/components/Posts.js
+++ b/examples/create-react-app/src/components/Posts.js
@@ -1,7 +1,7 @@
 import { useQuery } from 'graphql-hooks'
 import T from 'prop-types'
 import React from 'react'
-import CreatePost from './CreatePost'
+import CreatePost, { createPostMutation } from './CreatePost'
 
 export const allPostsQuery = `
   query {
@@ -14,7 +14,13 @@ export const allPostsQuery = `
 `
 
 export default function Posts() {
-  const { loading, data, error, refetch } = useQuery(allPostsQuery)
+  const { loading, data, error, refetch } = useQuery(allPostsQuery, {
+    refetchAfterMutations: [
+      {
+        mutation: createPostMutation
+      }
+    ]
+  })
 
   return (
     <>
--- a/examples/create-react-app/src/components/Posts.js
+++ b/examples/create-react-app/src/components/Posts.js
@@ -15,11 +15,7 @@ export const allPostsQuery = `
 
 export default function Posts() {
   const { loading, data, error, refetch } = useQuery(allPostsQuery, {
-    refetchAfterMutations: [
-      {
-        mutation: createPostMutation
-      }
-    ]
+    refetchAfterMutations: createPostMutation
   })
 
   return (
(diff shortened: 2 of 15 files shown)


giacomorebonato:
I believe **deploy-preview** step is failing because package-lock is still generated with node 14.
I'll address that too in this PR.

packages/graphql-hooks/package.json
```diff
@@ -35,6 +35,7 @@
   },
   "dependencies": {
     "dequal": "^2.0.0",
+    "events": "^3.3.0",
```
allforabit, line 38: Will this new dependency have an impact on the bundle size?

packages/graphql-hooks/index.d.ts
```diff
@@ -183,10 +187,16 @@ export interface UseClientRequestOptions<ResponseData = any, Variables = object>
...
+type RefetchAferMutationsData = {
+  mutation: string
+  filter?: (variables: any) => void
+}
+
 export interface UseQueryOptions<ResponseData = any, Variables = object>
   extends UseClientRequestOptions<ResponseData, Variables> {
   ssr?: boolean
   skip?: boolean
+  refetchAfterMutations?: RefetchAferMutationsData
```
allforabit, line 199: Is this type signature correct? The usage seems to suggest that it's an array.

allforabit (changes requested):

allforabit:
I've left a couple of comments. I want to try it out on techbase but just wanted to ensure I'm using it correctly as the type signature seems to be wrong. I'm assuming it should be an array of mutation objects. I'm wondering if it needs to be an array of objects and could be optionally an array of strings with just the mutations?

packages/graphql-hooks/package.json
giacomorebonato: Screenshot 2021-11-18 at 12 00 21

Minimal, but there is. Do you believe that we are still targeting the goal of this project of having a minimal bundle size?

simoneb: looks good

packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -1,5 +1,5 @@
-import React from 'react'
 import { dequal } from 'dequal'
+import React from 'react'
```
simoneb, line 2: I wonder how this would work before without this import!

packages/graphql-hooks/index.d.ts
giacomorebonato: Yeah, it's incorrect. Thanks, I'll fix it.

packages/graphql-hooks/package.json
allforabit: It does seem to be a selling point as far as I can see but quite new to the library/project so I may not be the most qualified person to say :-) I've seen people using a js map to track subscribers (I think it was in xState but I can't find it now.). Maybe this could be used if size is an issue.

simoneb (commented):
I didn't get a chance review thoroughly but upon a quick look this looks sensible and overall very good. 

Considering the large change I would suggest to:

1. try it out on a project before publishing a new version to make sure everything keeps working. @allforabit can do that
2. publish a new version as a prerelease then test it out more thoroughly. I can do this
3. once we're happy, release a final version

packages/graphql-hooks/package.json
simoneb: there isn't a strong argument in favor of this library but overall it looks like a reasonable compromise

packages/graphql-hooks/src/useClientRequest.js
giacomorebonato: I believe that my "fix on save" just swapped the order of the imports and the GitHub is not the diff properly.

examples/create-react-app/src/components/Posts.js
```diff
@@ -14,7 +14,13 @@ export const allPostsQuery = `
 `
 
 export default function Posts() {
-  const { loading, data, error, refetch } = useQuery(allPostsQuery)
+  const { loading, data, error, refetch } = useQuery(allPostsQuery, {
+    refetchAfterMutations: [
+      {
+        mutation: createPostMutation
```
simoneb, line 20: Maybe I missed something, but I thought that the mutation we're specifying here is the **name** of the mutation, not the string of the mutation itself. Is it intentional that we are specifying the full body of the mutation instead of its name?

Asking especially because I can surely see the advantage in doing so (you don't have to hardcode names, since they in the application they have little relevance), but on the other hand I wonder if using as key the whole body of a mutation is a robust approach.

README.md
```diff
@@ -243,6 +243,9 @@ This is a custom hook that takes care of fetching your query and storing the res
     - `previousData`: Previous GraphQL query or `updateData` result
     - `data`: New GraphQL query result
   - `client`: GraphQLClient - If a GraphQLClient is explicitly passed as an option, then it will be used instead of the client from the `ClientContext`.
+  - `refetchAfterMutations`: Array of RefetchAferMutationsData - Mutations that will cause the refetch of the query
+    - `mutation`: String - The mutation string
+    - `filter`: Function (optional) - It receives mutation variables as parameter and blocks refetch if it returns false
```
simoneb, line 248: let's make this option "dynamic" to be more user friendly and accept:

- string
- object
- array of {string | object}

the code will have to change to handle those cases

examples/typescript/src/App.tsx
```diff
@@ -20,7 +20,7 @@ const client = new GraphQLClient({
 
 export const allPostsQuery = `
   query {
-    allPosts(orderBy: createdAt_DESC, first: 20) {
```
simoneb, line 23: did you remove the order and limit because they're not needed, or?

packages/graphql-hooks/index.d.ts
```diff
@@ -183,10 +187,16 @@ export interface UseClientRequestOptions<ResponseData = any, Variables = object>
   client?: GraphQLClient
 }
 
+type RefetchAferMutationsData = {
+  mutation: string
+  filter?: (variables: any) => boolean
```
simoneb, line 192: variables is going to be an object I guess?

examples/typescript/src/App.tsx
giacomorebonato: I removed them because they are not used by the example and they break if used with the included `json-graphql-server` which doesn't implement those parameters.

simoneb: I thought we were already using json-graphql-server, didn't it work?

giacomorebonato: This example was broken when using the local json-graphql-server for this reason, but working when pointing to the remote server.

packages/graphql-hooks/src/useQuery.js
```diff
@@ -46,20 +45,57 @@ function useQuery(query, opts = {}) {
...
+        }
+
+        return acc
+      }, {})
+
+      const conditionalRefetch = ({ mutation, variables }) => {
+        const { filter } = mutationsMap[mutation]
+
+        if (!filter || (variables && filter(variables))) {
+          refetch()
```
simoneb, line 79: how do we handle the case when a component unmounts in the middle of a refetch? won't we have to handle the error here or is it handled already by this library somewhere else? iow it's the typical problem that you trigger an async operation which will need to update local state, and while the operation is in progress the component unmounts. if you try to update the state of an unmounted component you'll get a React error. 
here we don't have access to the state because it's stored elsewhere, but I wonder if we need to catch and swallow the error instead

giacomorebonato: This makes me think about this change in React:
* https://github.com/facebook/react/pull/22114

I need to think if our case is different and needs special adjustments, but it might not.

examples/create-react-app/src/components/Posts.js
giacomorebonato: Yes, it's intentionally the entire mutation string.
I thought of that because it's what we already have handy for matching with a mutation and avoiding mistakes.
Other ideas are welcome :)

packages/graphql-hooks/src/useQuery.js
simoneb: Ah very interesting, I wasn't aware of this! I see it's merged, do we know if it's released too? I'm fairly sure I've seen this error still occurring recently

examples/create-react-app/src/components/Posts.js
simoneb: sounds good, let's make sure this is clear in the docs because it's not obvious

packages/graphql-hooks/src/useQuery.js
giacomorebonato: It is not released yet. The PR is from August while the last release of React was in March.
* I think that we shouldn't manage this situation though and treat the error message as an expected false positive
* the message is useful in case we forget to clean up the subscription (but we didn't forget about that)
* React is using `console.error` and not throwing an Error, so it wouldn't help to "catch"

Please let me know if you have different thoughts about this.

simoneb: sounds perfectly reasonable

README.md
```diff
@@ -243,6 +243,11 @@ This is a custom hook that takes care of fetching your query and storing the res
     - `previousData`: Previous GraphQL query or `updateData` result
     - `data`: New GraphQL query result
   - `client`: GraphQLClient - If a GraphQLClient is explicitly passed as an option, then it will be used instead of the client from the `ClientContext`.
+  - `refetchAfterMutations`: list of mutations that will trigger the query refetch. It can be either:
+    - String | String[] - The string containing the mutation's definition or an array of them
+    - Object | Object[] with the following properties or an array of them
+      - `mutation`: String - The mutation string
+      - `filter`: Function (optional) - It receives mutation variables as parameter and blocks refetch if it returns false
```
simoneb, line 250: I would consider describing it in a different way:

- avoid mentioning "list of mutations", because in fact it could be only one
- the type definition should actually be `string | object | (string | object)[]`. Then, once you define the type you describe
  - if it's a string, it's the mutation string
  - if it's an object then it has properties mutation and filter
  - if it's an array, the elements can be of either type above

rephrase in a way that makes sense to the reader :)

packages/graphql-hooks/index.d.ts
```diff
@@ -183,10 +187,20 @@ export interface UseClientRequestOptions<ResponseData = any, Variables = object>
...
 export interface UseQueryOptions<ResponseData = any, Variables = object>
   extends UseClientRequestOptions<ResponseData, Variables> {
   ssr?: boolean
   skip?: boolean
+  refetchAfterMutations?:
+    | string
+    | string[]
+    | RefetchAferMutationsData
+    | RefetchAferMutationsData[]
 }
```
simoneb, line 204: why don't we define the RefetchAfterMutationsData as a union type which represents all the possible options instead of doing it here?

packages/graphql-hooks/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "graphql-hooks",
-  "version": "5.4.1",
+  "version": "5.5.0",
```
simoneb, line 3: don't bump version numbers in PRs, they're bumped when we release

packages/graphql-hooks/src/createRefetchMutationsMap.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * Checks values of refetchAfterMutations public option and maps them to an object
+ * @typedef {import('../index').RefetchAferMutationsData} RefetchAferMutationsData
+ *
+ * @param {string | string[] | RefetchAferMutationsData | RefetchAferMutationsData[]} refetchAfterMutations
```
simoneb, line 5: the comment above would simplify this, too

simoneb (approved):

PR #688, opened 2021-11-18
fix: update to node 16

giacomorebonato:
### What does this PR do?

Updates the project to be built with Node 16.
Because it's LTS and also because Netlify requires it in its build step.

### Checklist

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -14,17 +14,13 @@ jobs:
   acceptance-tests:
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        node-version: [16.x]
-
     steps:
       - run: echo Running on target url $ACCEPTANCE_URL
       - uses: actions/checkout@v2.3.4
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Use Node.js 16
         uses: actions/setup-node@v2.4.1
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: 16
       - name: Run tests
         uses: DevExpress/testcafe-action@latest
         with:
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [16.x]
 
     steps:
       - run: echo Running on target url $ACCEPTANCE_URL
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -17,8 +17,7 @@ jobs:
     steps:
       - run: echo Running on target url $ACCEPTANCE_URL
       - uses: actions/checkout@v2.3.4
-      - name: Use Node.js 16
-        uses: actions/setup-node@v2.4.1
+      - uses: actions/setup-node@v2.4.1
         with:
           node-version: 16
       - name: Run tests
--- a/examples/fastify-ssr/package.json
+++ b/examples/fastify-ssr/package.json
@@ -20,7 +20,6 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@reach/router": "^1.2.1",
     "babel-plugin-dynamic-import-node": "^2.2.0",
     "fastify": "^3.0.0",
     "fastify-gql": "^5.0.0",
@@ -29,9 +28,11 @@
     "graphql-hooks-memcache": "^2.2.1",
     "graphql-hooks-ssr": "^2.2.1",
     "isomorphic-unfetch": "^3.0.0",
+    "prop-types": "^15.7.2",
     "react": "^17.0.0",
     "react-dom": "^17.0.0",
-    "react-tree-walker": "^4.3.0"
+    "react-router": "^6.0.2",
+    "react-router-dom": "^6.0.2"
   },
   "devDependencies": {
     "@babel/cli": "^7.4.4",
--- a/examples/fastify-ssr/package.json
+++ b/examples/fastify-ssr/package.json
@@ -31,7 +31,6 @@
     "prop-types": "^15.7.2",
     "react": "^17.0.0",
     "react-dom": "^17.0.0",
-    "react-router": "^6.0.2",
     "react-router-dom": "^6.0.2"
   },
   "devDependencies": {
--- a/examples/fastify-ssr/src/app/AppShell.js
+++ b/examples/fastify-ssr/src/app/AppShell.js
@@ -5,7 +5,6 @@ import HomePage from './pages/HomePage'
 import NotFoundPage from './pages/NotFoundPage'
 import PaginationPage from './pages/PaginationPage'
 
-
 class AppShell extends React.Component {
   render() {
     return (
--- a/examples/fastify-ssr/src/app/AppShell.js
+++ b/examples/fastify-ssr/src/app/AppShell.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Link, Router } from 'react-router-dom'
+import { Link } from 'react-router-dom'
 import HomePage from './pages/HomePage'
 // components
 import NotFoundPage from './pages/NotFoundPage'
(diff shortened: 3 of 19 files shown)


simoneb:
This PR seems to contain other things, can you please describe all changes?

examples/fastify-ssr/package.json
```diff
@@ -20,7 +20,6 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@reach/router": "^1.2.1",
```
giacomorebonato, line 23: `npm i` was failing because @reach/router ([even in its latest version](https://github.com/reach/router/blob/master/package.json#L29)) has a peer dependency of React < 17

examples/fastify-ssr/package.json
```diff
@@ -29,9 +28,11 @@
     "graphql-hooks-memcache": "^2.2.1",
     "graphql-hooks-ssr": "^2.2.1",
     "isomorphic-unfetch": "^3.0.0",
+    "prop-types": "^15.7.2",
     "react": "^17.0.0",
     "react-dom": "^17.0.0",
-    "react-tree-walker": "^4.3.0"
```
giacomorebonato, line 34: This also has peer [dependency of React < 17](https://github.com/ctrlplusb/react-tree-walker/blob/master/package.json#L36), but it's also unused in this sample project. That's the reason I removed it.

.github/workflows/acceptance-tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [16.x]
```
simoneb, line 19: let's remove matrixes altogether if we're only ever using a single node version. let's provide the node version in the setup-node action inputs, and let's also use 16 instead of 16.x

examples/fastify-ssr/package.json
```diff
@@ -29,9 +28,11 @@
...
+    "react-router": "^6.0.2",
```
simoneb, line 34: react-router is not necessary anymore I believe (it used to in earlier versions)

examples/fastify-ssr/src/server/handlers/app-shell.js
```diff
@@ -1,6 +1,6 @@
 const React = require('react')
 const ReactDOMServer = require('react-dom/server')
-const { ServerLocation } = require('@reach/router')
+const { StaticRouter } = require('react-router')
```
giacomorebonato, line 3: @simoneb react-router is used here, but I don't know if it's effectively needed.

simoneb: I don't know, can you please check the docs? https://reactrouter.com/docs/en/v6/guides/ssr

giacomorebonato: Thanks, good advice. It's not needed indeed, I'll push the changes.

examples/typescript/package.json
```diff
@@ -12,7 +12,7 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-scripts": "^3.4.4",
-    "typescript": "^4.0.5"
+    "typescript": "^3.2.1"
```
giacomorebonato, line 15: Needed because it's "react-scripts": "^3.4.4" peer dependency

.github/workflows/acceptance-tests.yml
```diff
@@ -14,17 +14,13 @@ jobs:
...
 
-    strategy:
-      matrix:
-        node-version: [14.x]
-
     steps:
       - run: echo Running on target url $ACCEPTANCE_URL
       - uses: actions/checkout@v2.3.4
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Use Node.js 16
```
simoneb, line 20: nit: you can remove these names, they're fairly redundant and what the action is doing is quite clear from the name itself

examples/fastify-ssr/src/client/js/app-shell.js
```diff
@@ -1,22 +1,23 @@
+// graphql-hooks
```
simoneb, line 1: let's remove this comment, doesn't seem useful (I realize it was already there)

examples/typescript/package.json
simoneb: dependabot will bump this dependency if we downgrade it manually though

simoneb (approved):
A few comments but overall looks good to me

giacomorebonato:
Thanks @simoneb , I am addressing the comments. 
And check why I broke again the build.

packages/graphql-hooks/package.json
```diff
@@ -48,7 +48,6 @@
     "react": "^17.0.0",
     "react-dom": "^17.0.0",
     "react-test-renderer": "^17.0.0",
-    "rollup": "^2.0.0",
```
giacomorebonato, line 51: Screenshot 2021-11-18 at 12 12 30

This is mentioned among other Lerna warnings, but it needed to be addressed.
It was somehow causing this documented issue: https://reactjs.org/warnings/invalid-hook-call-warning.html
Probably by generating some mismatching version in the bundle.

packages/graphql-hooks/test/unit/useClientRequest.test.js
```diff
@@ -1,7 +1,7 @@
-import React from 'react'
+import { act, renderHook } from '@testing-library/react-hooks'
```
giacomorebonato, line 1: This is just formatting changes.

packages/graphql-hooks-memcache/package.json
```diff
@@ -40,8 +40,5 @@
   "bugs": {
     "url": "https://github.com/nearform/graphql-hooks/issues"
   },
-  "homepage": "https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks-memcache#readme",
-  "devDependencies": {
-    "rollup": "^2.0.0"
```
giacomorebonato, line 45: https://github.com/nearform/graphql-hooks/pull/688/files#r752186663

examples/fastify-ssr/src/app/AppShell.js
```diff
@@ -15,11 +14,10 @@ class AppShell extends React.Component {
           <Link to="/">Home</Link>
           <Link to="/users">PaginationPage</Link>
         </nav>
-        <Router>
```
giacomorebonato, line 18: Router moved to parent component, so this one can be imported on server side, which uses a different kind of router.

examples/typescript/package.json
```diff
@@ -5,14 +5,14 @@
...
     "@types/react-dom": "^16.9.9",
     "graphql-hooks": "^5.4.1",
     "graphql-hooks-memcache": "^2.2.1",
-    "react": "^17.0.1",
-    "react-dom": "^17.0.1",
-    "react-scripts": "^3.4.4",
-    "typescript": "^4.0.5"
+    "react": "^17.0.0",
+    "react-dom": "^17.0.0",
+    "react-scripts": "^4.0.3",
```
giacomorebonato, line 14: Upgrading react-scripts allows an higher peer dependency of TypeScript.
Both the examples with create-react-app are compiling and working.

examples/typescript/src/App.tsx
```diff
@@ -1,14 +1,14 @@
...
+  GraphQLClient,
   useManualQuery,
-  APIError
+  useMutation,
+  useQuery
 } from 'graphql-hooks'
+import React, { useState } from 'react'
 
-interface Post {
+interface PostData {
```
giacomorebonato, line 11: This renaming was needed to avoid an eslint warning (but maybe error on CI).
All the rest is formatting.

package.json
```diff
@@ -5,7 +5,7 @@
     "test": "jest",
     "test:coverage": "jest --coverage",
     "test:acceptance": "testcafe chrome test/acceptance/",
-    "postinstall": "lerna bootstrap --no-ci && lerna run build",
+    "postinstall": "lerna bootstrap --hoist --no-ci && lerna run build",
```
giacomorebonato, line 8: This is the main change that make things work.
`lerna bootstrap` behaves differently with this version of Node and without the `--hoist` parameter it doesn't save anything in the root `./node_modules/.bin`, meaning that without this change also the sequent `lerna run build` fails straight away because of "command not found".

giacomorebonato:
All comments addressed. @simoneb if you want to double checked I tried to include only necessary changes.

simoneb (approved):
Let's add or update the engine entry in top level package file to require node 16

Issue #729, opened 2021-12-13
Response headers


zanettin:
Hi 👋 

First of all - thanks a lot for this neat pice of OSS 🙏 

### Description
The current implementation does "just" return the fetched data or errors from the fetch itself. We would be interested in the response headers since we have to forward the cache-tags to the SSR generated pages so we can cache and purge it properly on our CDN.

### Suggested implementation
Just checked the src code of the client real quick and the simplest way to achieve that would be to extend the `generateResult` fn and return the headers as a new key. a more generic solution would be to pass an (array) of middleware fns as option into the `request` method (`{middleware: [enrichData(resp)]}`) so the user can define what data he/she is interested from the original fetch response. the real problem would be the caching since we would need to cache the headers as well (at least for non SSR requests). not yet digged into that.

did you already have thoughts in this direction? i'd like to implement smth but would like to hear your thoughts and desired path first 😄

PR #730, opened 2021-12-13
feat: added response proxy

zanettin:
### What does this PR do?
It's an implementation **proposal** for this issue https://github.com/nearform/graphql-hooks/issues/729. The goal is to give the user access to all properties on the Fetch API `Response` object. In our case we're interested in the headers to get the assigned cache keys on SSR rendered apps. With this approach, the user can define in which properties he/she is interested and just those will be returned under the defined key on the response of `client.request()` calls.

### Related issues
https://github.com/nearform/graphql-hooks/issues/729

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ const client = new GraphQLClient(config)
 - `request(operation, options)`: Make a request to your GraphQL server; returning a Promise
   - `operation`: Object with `query`, `variables` and `operationName`
 - `options.fetchOptionsOverrides`: Object containing additional fetch options to be added to the default ones passed to `new GraphQLClient(config)`
-- `options.responseReducer`: Reducer function to pick values from the original Fetch Response object. Values are appended to the `request` response under the `responseReducer` key. Example usage: `{responseReducer: (response) => ({myKey: response.headers.get('content-length)})`
+- `options.responseReducer`: Reducer function to pick values from the original Fetch Response object. Values are merged to the `request` response under the `data` key. Example usage: `{responseReducer: (data, response) => ({...data, myKey: response.headers.get('content-length)})`
 
 ## `ClientContext`
 
--- a/README.md
+++ b/README.md
@@ -172,6 +172,7 @@ const client = new GraphQLClient(config)
 - `request(operation, options)`: Make a request to your GraphQL server; returning a Promise
   - `operation`: Object with `query`, `variables` and `operationName`
 - `options.fetchOptionsOverrides`: Object containing additional fetch options to be added to the default ones passed to `new GraphQLClient(config)`
+- `options.responseProxy`: Key-Value pairs to map parts of the original Fetch API response into the return value. Example for header forwarding would look like `{responseProxy: {header: resp => resp.headers}}`. Accessable is everything which the chosen Fetch implementation supports on the `Response` object.
 
 ## `ClientContext`
 
@@ -615,12 +616,11 @@ useQuery(allPostsByUserIdQuery, {
   refetchAfterMutations: [
     {
       mutation: createPostMutation,
-      filter: (variables) => variables.userId === myUserId
+      filter: variables => variables.userId === myUserId
     }
   ]
 })
 ```
-
 
 ## File uploads
 
@@ -993,6 +993,7 @@ const client = new GraphQLClient({
 
 We now use GitHub Discussions for our community. To join, click on ["Discussions"](https://github.com/nearform/graphql-hooks/discussions). We encourage you to start a new discussion, share some ideas or ask questions from the community.
 If you want to see the old community posts (on Spectrum) you can access them [here](https://spectrum.chat/graphql-hooks).
+
 ## Contributors
 
 Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/docs/en/emoji-key)):
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ const client = new GraphQLClient(config)
 - `request(operation, options)`: Make a request to your GraphQL server; returning a Promise
   - `operation`: Object with `query`, `variables` and `operationName`
 - `options.fetchOptionsOverrides`: Object containing additional fetch options to be added to the default ones passed to `new GraphQLClient(config)`
-- `options.responseProxy`: Key-Value pairs to map parts of the original Fetch API response into the return value. Example for header forwarding would look like `{responseProxy: {header: resp => resp.headers}}`. Accessable is everything which the chosen Fetch implementation supports on the `Response` object.
+- `options.responseReducer`: Reducer function to pick values from the original Fetch Response object. Values are appended to the `request` response under the `responseReducer` key. Example usage: `{responseReducer: (response) => ({myKey: response.header.get('content-length)})`
 
 ## `ClientContext`
 
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ const client = new GraphQLClient(config)
 - `request(operation, options)`: Make a request to your GraphQL server; returning a Promise
   - `operation`: Object with `query`, `variables` and `operationName`
 - `options.fetchOptionsOverrides`: Object containing additional fetch options to be added to the default ones passed to `new GraphQLClient(config)`
-- `options.responseReducer`: Reducer function to pick values from the original Fetch Response object. Values are appended to the `request` response under the `responseReducer` key. Example usage: `{responseReducer: (response) => ({myKey: response.header.get('content-length)})`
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -172,6 +172,7 @@ const client = new GraphQLClient(config)
 - `request(operation, options)`: Make a request to your GraphQL server; returning a Promise
   - `operation`: Object with `query`, `variables` and `operationName`
 - `options.fetchOptionsOverrides`: Object containing additional fetch options to be added to the default ones passed to `new GraphQLClient(config)`
+- `options.responseProxy`: Key-Value pairs to map parts of the original Fetch API response into the return value. Example for header forwarding would look like `{responseProxy: {header: resp => resp.headers}}`. Accessable is everything which the chosen Fetch implementation supports on the `Response` object.
```
gdorsi, line 175: Hello!

What do you think of exposing a simple serialize function instead?

I'm thinking about something like this:
```js
{ responseReducer: (response) => ({ cookies: response.headers.get('cookies') }) }
```

This solution would be more performant, because we pick only what's needed.

packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -90,14 +90,20 @@ class GraphQLClient {
...
+    responseProxy = {}
+  }) {
     const errorFound = !!(
       (graphQLErrors && graphQLErrors.length > 0) ||
       fetchError ||
       httpError
     )
     return !errorFound
-      ? { data }
+      ? { ...responseProxy, data }
```
gdorsi, line 106: I think that would be better to append the result to a new key (e.g. `responseData`) instead of merging

It would be easier to add typescript types this way.

README.md
zanettin: thanks for your feedbacks @gdorsi 🙏  I've updated the implementation and used a reducer instead.

packages/graphql-hooks/src/GraphQLClient.js
zanettin: makes sene 👍  adjusted that as wellw

simoneb:
can you fix the build?

packages/graphql-hooks/index.d.ts
```diff
@@ -185,6 +185,7 @@ export interface UseClientRequestOptions<
   fetchOptionsOverrides?: object
   updateData?(previousData: ResponseData, data: ResponseData): any
   client?: GraphQLClient
+  responseReducer?(response: object): object
```
gdorsi, line 188: We can be more specific here:

```diff
- responseReducer?(response: object): object
+ responseReducer?(response: Response): object
```

The `UseClientRequestResult` type also needs to be updated with the new key.

packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -90,14 +90,20 @@ class GraphQLClient {
   }
   /* eslint-enable no-console */
 
-  generateResult({ fetchError, httpError, graphQLErrors, data }) {
+  generateResult({
+    fetchError,
+    httpError,
+    graphQLErrors,
+    data,
+    responseReducer = null
```
gdorsi, line 98: Is the default value necessary here?

zanettin:
@simoneb just run it again locally and the build just went through:
```
The build folder is ready to be deployed.
You may serve it with a static server:

  npm install -g serve
  serve -s build

Find out more about deployment here:

  bit.ly/CRA-deploy

lerna success run Ran npm script 'build' in 7 packages in 8.6s:
lerna success - create-react-app
lerna success - fastify-ssr
lerna success - subscription
lerna success - typescript-example
lerna success - babel-plugin-extract-gql
lerna success - graphql-hooks-memcache
lerna success - graphql-hooks
✨  Done in 10.25s.
```
anything lerna/npm related?

packages/graphql-hooks/index.d.ts
zanettin: Hi @gdorsi thanks again. Does `Response` refer to the Fetch API? if so we shouldn't change it since graphql-hooks allow to provide other fetch implementations which then may have another signature, right?

packages/graphql-hooks/src/GraphQLClient.js
zanettin: no, not needed. just a habit of mine i'd say. can remove it if needed 👍

packages/graphql-hooks/index.d.ts
gdorsi: uh, thats true!

You have done right XD

packages/graphql-hooks/src/GraphQLClient.js
gdorsi: Not a big deal, i only think that removing it would make the style more consinstent with the rest

zanettin:
@simoneb just seen that the latest master failed as well?
https://github.com/nearform/graphql-hooks/actions/runs/1581934844

gdorsi:
I've also tried the build locally and everything works.

Maybe it's something caused by the node_modules cache?

packages/graphql-hooks/index.d.ts
zanettin: Hi @gdorsi 👋 
Yes just had a look at it and from MPOV we should already be fine with the current state.
The `request` fns uses generics as well but defaults to `any` if not specified 👍 
Bildschirmfoto 2021-12-16 um 10 38 00

so i made a quick check again on the TS example in the repo and passed in a custom generic for the responseData (resolve case of the promise):
Bildschirmfoto 2021-12-16 um 10 45 36

or did you mean smth different?
thanks again 🙏

gdorsi: I'm talking about this:

Screenshot 2021-12-16 at 16 50 10

Or I'm doing something wrong?

zanettin: Hi @gdorsi 
Thanks once again 🙏  now i understand what you've meant 😅 

Guess i'd have two options to solve this:
- return the `responseReducer` data within the `data` key of the response under an unique key (`__responseReducer` for example) to avoid collisions with fetched data. the generic in this case could look smth like this then: `client.request<{data: {some: string, cool: string, stuff: any}, __responseReducer: { contentType: string}}, any, any>`
- add a 4th generic to all methods which return `Result<ResponseData, TGraphqlError>` so that we can type everything there. example could look like `client.request<{data: {some: string, cool: string, stuff: any}}, any, any, { contentType: string, contentLength: number}>`

or do you have any other in mind? which one would you prefer?

<hr />

**random question**
when i've played around with the 2nd version i've realized that as soon as i change the `ResponseData` type from `any` to a specific value, the `data` key is set optional so i would have to use `data?.myValue` to access it. 

Bildschirmfoto 2021-12-17 um 22 11 47

Comes from L175 in the `index.d.ts` file. 
Bildschirmfoto 2021-12-17 um 22 11 31

How do you deal with that when you specify the data returned from the server? check if an error occurred and if not, cast the response before using it?
Bildschirmfoto 2021-12-17 um 22 28 33

gdorsi: > or do you have any other in mind? which one would you prefer?

I'm wondering, what if we change the behavoir of `responseReducer` and let the user enrich the `data` field with some information from the response?

Something like this:
```js
const { data } = await client.request({ query }, {
 responseReducer: (response, data) => ({
   ...data,
   cache: response.headers.get('x-cache-tags')
 })
})
```

What do you think?

zanettin: thank you @gdorsi 👍  i've implemented that one now

gdorsi (approved):
lgtm

simoneb:
Wouldn't a simpler implementation just return a `response` property containing the whole raw response alongside the existing `data`, `error` and `loading`?

```js
const { response, data, loading, error } = useQuery(...)
```

gdorsi:
> Wouldn't a simpler implementation just return a `response` property containing the whole raw response alongside the existing `data`, `error` and `loading`?
> 
> ```js
> const { response, data, loading, error } = useQuery(...)
> ```

Would this work with SSR? Or there is no need for SSR integration?

simoneb:
Why would it not work?

gdorsi:
Isn't the response object not serializable? 

But maybe I'm missing something XD

simoneb:
> Isn't the response object not serializable?
> 
> But maybe I'm missing something XD

Ah that's a good point. I guess it depends on how you serialize it!

simoneb (approved):
Okay let's merge this PR then, as we're not sure that sending back the whole response is a sensible default alternative to this implementation.

zanettin:
@simoneb and @gdorsi thanks again for your inputs and for the merge 🙏

PR #733, opened 2021-12-15
chore: fixes the build job on the ci workflow

gdorsi:
fixes https://github.com/nearform/graphql-hooks/issues/734

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
           key: npm-${{ hashFiles('**/package.json') }}
       - uses: actions/setup-node@v2
         with:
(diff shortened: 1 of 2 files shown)


gdorsi:
Pinning the node version to the v16.13.0 has fixed the issue.

I've reproduced the problem locally and the error is generated by `mini-css-webpack-lugin`, probably caused by how the dependencies are hoisted.

Needs more investigation, maybe we could try to move to the npm workspaces and see if it works there?

simoneb (approved):
We'll need to remember to go back and undo this change at some point. Can you create an issue and a PR which undoes this change? When that passes, we're good to revert

PR #751, opened 2021-12-28
feat: a mock client for graphql-hooks

### LINKED ISSUE #746: Add a Testing section in the docs
### Description

I think it would be very useful for newcomers to have a testing section where we give some examples of how to write good tests for applications based on graphql-hooks

### Suggested implementation

[MSW](https://mswjs.io/docs/getting-started/mocks/graphql-api) is a great mocking tool, I think that an example made with it and [react-testing-library](https://testing-library.com/docs/react-testing-library/example-intro) would be a good start.

gdorsi:
### What does this PR do?

The intention is to simplify the testing experience and speedup develompement by providing an easy way to mock the graphql responses.

### Related issues

closes #746 

### TODO

- [ ] Write the tests
- [ ] Write the docs
- [ ] Write some examples on how to define "global" mocks

### DIFF
--- a/examples/create-react-app/package.json
+++ b/examples/create-react-app/package.json
@@ -19,7 +19,6 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test",
     "eject": "react-scripts eject",
     "start:server": "json-graphql-server server/db.cjs --h 0.0.0.0 --p 8080"
   },
--- a/examples/create-react-app/test/CreatePost.test.js
+++ b/examples/create-react-app/test/CreatePost.test.js
@@ -1,0 +1,56 @@
+import { render, screen, fireEvent, waitFor } from './test-utils'
+import CreatePost, { createPostMutation } from '../src/components/CreatePost'
+import React from 'react'
+
+const localQueries = {
+  [createPostMutation]: () => ({ createPost: { id: 1 } })
+}
+
+describe('CreatePost', () => {
+  afterEach(() => {
+    jest.resetAllMocks()
+  })
+
+  it('should submit the new post', async () => {
+    const createPostSpy = jest.spyOn(localQueries, createPostMutation)
+
+    render(<CreatePost />, {
+      localQueries
+    })
+
+    fireEvent.input(
+      screen.getByRole('textbox', {
+        name: /title/i
+      }),
+      {
+        target: {
+          value: 'Test'
+        }
+      }
+    )
+
+    fireEvent.input(
+      screen.getByRole('textbox', {
+        name: /url/i
+      }),
+      {
+        target: {
+          value: 'https://example.com'
+        }
+      }
+    )
+
+    fireEvent.click(
+      screen.getByRole('button', {
+        name: /Add post/i
+      })
+    )
+
+    waitFor(() =>
+      expect(createPostSpy).toHaveBeenCalledWith({
+        title: 'Test',
+        url: 'https://example.com'
+      })
+    )
+  })
+})
--- a/examples/create-react-app/test/Posts.test.js
+++ b/examples/create-react-app/test/Posts.test.js
@@ -1,0 +1,29 @@
+import { render, screen } from './test-utils'
+import Posts, { allPostsQuery } from '../src/components/Posts'
+import React from 'react'
+
+const localQueries = {
+  [allPostsQuery]: () => ({
+    allPosts: [
+      {
+        id: 1,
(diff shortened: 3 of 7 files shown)


packages/graphql-hooks-mock-client/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "hooks",
+    "react",
+    "graphql-hooks",
+    "cache",
+    "mocks",
+    "mock client"
+  ],
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "graphql-hooks": "^5.5.0"
```
gdorsi, line 23: I though that a `peerDependency` would be the best choice but `lerna bootstrap` doesn't automatically link `graphql-hooks` in the `node_modules`.

I've made a workaround by manually creating the link inside the `packages` dir, but maybe it would be easier to move the `graphql-hooks` inside the dependencies.
That would also enable the automatic version bump when we do a change on `graphql-hooks`.

Does someone know what is the best pratice here?

gdorsi:
I've not updated the main `README` because I would like to test this library in some projects before claiming it as best pratice for testing.

packages/graphql-hooks-mock-client/src/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+      if (mock) {
+        const result = mock(
+          { query: operation.query, variables: operation.variables },
+          options
+        )
+
+        return Promise.resolve(result)
+      }
+    }
```
mcollina, line 85: you are essentially reimplementing GraphQL - without most of the edge cases. I think a more scalable approach could be to have a "LocalClient" that create a full schema, set up some resolvers and then calls https://graphql.org/graphql-js/graphql/#graphql. It might be useful on more situations.

gdorsi: hmm interesting...

So basically, instead of returning the data as-is we could create the schema/resolvers on-the-fly based on the mock return value and run the request against it.

This way we get the graphql server features without making the mocks more complicated.

I'll try to hack something and see how far I can go.

Thanks for the tip!

simoneb (commented):
I like the idea you came up with here @gdorsi but I agree with Matteo that we should probably try to take this one step further and avoid reimplementing graphql. I see two options here:

1. we follow something similar to what Matteo suggested but try to simplify it, in the sense that we probably don't want to have to specify a full schema, since this is a client side library and we don't really care about it. I agree that it would make setting up a mock client overly and unnecessarily complex. We could also inspire from how apollo does it: https://www.apollographql.com/docs/react/development-testing/client-schema-mocking/
2. we could build this on top of MSW, basically providing a higher-level API on top of it https://mswjs.io/docs/getting-started/mocks/graphql-api

If none of those two options works, then I'm actually fine with this solution as well. In any case very good job at trying to come up with a solution, which I agree we do need at some point as several people over time have asked to have some guides for testing it.

gdorsi:
Thanks for all your feedback!

> We could also inspire from how apollo does it

Thanks for sharing it, very inspiring!

The Apollo mocking system is near where I want to be. 
I'd like to make an API that could improve velocity in the first phases of development by giving the capability of mocking parts of the backend and also simplify testing providing an easy way to mock the backand responses.

> we could build this on top of MSW, basically providing a higher-level API on top of it

My initial goal was to provide the same MSW capabilities without relying on [monkey patching](https://github.com/mswjs/interceptors/blob/main/src/interceptors/ClientRequest/index.ts#L24).
So I would prefer to not use it under the hood.

Summing up my goals are:
- Improve the testing experience for `graphql-hooks`
- Provide an API for the partial mocking of the backend 

All by trying to follow the graphql standard as much as possible.

The constraints i would like to set are:
- Don't rely on monkey patching/global pollution
- Don't bloat the main library, this should be an optional feature

gdorsi:
I've updated the implementation taking inspiration on how [urlql](https://formidable.com/open-source/urql/docs/advanced/testing/) handles testing.

To show how this works I've added some tests in the CRA example project.

This new implementation is also open to evolve into what @mcollina suggests with something like this:

```js
    local: new LocalResolver(opts)
  })
```

gdorsi:
> I'm wondering if it would be easier to create different LocalGraphQLClient class that inherits from the first one and overrides methods like request as needed with different logic.

mmhh seems a good idea.

Also, if we implement this by extension, this feature becomes three-shackable.
We could even move it in a different package.

gdorsi:
Closing in favor of https://github.com/nearform/graphql-hooks/pull/797

Issue #681, opened 2021-11-05
`Error: WebSocket implementation missing`, even with `ssrMode` disabled


bitjson:
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: `5.4.1`
- `graphql-ws` version: `5.5.5`
- `react` version: `17.0.2`

### Description

When using `graphql-hooks`+`graphql-ws` in a new Next.js app, `graphql-ws` emits the error `Error: WebSocket implementation missing` during server-side rendering, even when `ssrMode` is disabled:

```ts
// in _app.tsx:
const client = new GraphQLClient({
  url: '//localhost:8080/v1/graphql',
  ssrMode: false, // disabled by default, but set to false here to confirm
  subscriptionClient: createClient({
    url: 'ws://localhost:8080/v1/graphql',
  }),
});
```

I'm currently using this as a workaround, but it would be nice if `graphql-hooks` didn't error "out of the box" when `ssrMode` is disabled (and no websocket implementation is installed on the server):

```ts
// in _app.tsx:
const ssr = typeof window === 'undefined';
const client = new GraphQLClient({
  url: '//localhost:8080/v1/graphql',
  ssrMode: false,
  ...(ssr
    ? {}
    : {
        subscriptionClient: createClient({
          url: 'ws://localhost:8080/v1/graphql',
        }),
      }),
});
```

### How to reproduce

```sh
yarn create next-app --typescript
cd my-app
yarn add graphql graphql-hooks graphql-ws
```

Create the client in `_app.tsx`:
```ts
import '../styles/globals.css';
import type { AppProps } from 'next/app';
import { GraphQLClient, ClientContext } from 'graphql-hooks';
import { createClient } from 'graphql-ws';

const client = new GraphQLClient({
  url: '//localhost:8080/v1/graphql',
  subscriptionClient: createClient({
    url: 'ws://localhost:8080/v1/graphql',
  }),
});

function MyApp({ Component, pageProps }: AppProps) {
  return (
    <ClientContext.Provider value={client}>
      <Component {...pageProps} />
    </ClientContext.Provider>
  );
}

export default MyApp;
```

Then run `yarn dev`, and the error is produced:
```
Error: WebSocket implementation missing
    at createClient (file:///[path]/my-app/node_modules/graphql-ws/lib/client.mjs:63:15)
```

allforabit:
[...]

PR #755, opened 2021-12-29
feat: subscriptionClient accepts now a factory function as value

gdorsi:
### What does this PR do?

Adds an option for lazily instantiate the subscription client in order to provide an easier support for ssr without introducing breaking changes.

### Related issues

closes https://github.com/nearform/graphql-hooks/issues/681

### Checklist


mcollina (changes requested):
I would avoid adding a new option, but add support into `subscriptionClient` to functions, so that a function could be passed in or a `SubscriptionClient`.

simoneb (approved):

PR #767, opened 2022-01-18
feat: add fullWsTransport support

### LINKED ISSUE #766: Queries and Mutations via WebSocket
I'm adding support for this feature in `mercurius`: https://github.com/mercurius-js/mercurius/pull/705

Once it lands I think we should add the same support in `graphql-hooks`.

The API will not be changed, in fact this one should just work:

```jsx
const client = createClient({
  url: "ws://localhost:4000/graphql",
});
```

Now it doesn't, the client throws `'GraphQLClient: config.fetch must be a function'`.

After that, the hooks will be used as usual, no changes.

At first it seems easy to implement but I'm not sure if it would clash with some existing features.

nuragic:
Add support for sending queries and mutations via WS, as proposed in [this issue](https://github.com/nearform/graphql-hooks/issues/766).

Tests are passing and I haven't touched them, so I think it's a good sign because this is not intended to be a breaking change. 😄 

### Next steps
- [x] validate proposal
- [x] add tests
- [x] add an example app on `examples`
- [x] update docs

Close #766.


nuragic:
This branch can be tested with this client: https://github.com/nuragic/graphql-react-app/blob/graphql-hooks/

It involves some `npm link`_ing_ but it's fine 🔗 📦 🙃 

The only missing thing on that demo app is the manual cache update, but I opened another issue for that: #768

simoneb:
Overall looks good so far, what's missing in this PR?

simoneb:
Resolve the todo and write documentation, yes. Shall we have an example showing how to use this feature, too?

nuragic:
Ok, I'll also add an example.

krisgrint (commented):
Hi @nuragic just two tiny things with regard to the example app

nuragic:
@krisgrint Thanks for the review! Just fixed those right now.

krisgrint (approved):
LGTM!

hails (approved):

packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -189,6 +202,18 @@ class GraphQLClient {
   }
 
   request(operation, options = {}) {
+    if (this.fullWsTransport) {
+      return this.requestViaWS(operation)
+    } else {
```
simoneb, line 207: redundant else

packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -189,6 +202,18 @@ class GraphQLClient {
...
+      if (this.url) {
+        return this.requestViaHttp(operation, options)
+      } else {
```
simoneb, line 210: redundant else

simoneb (changes requested):
@hails a couple of comments, can you please address them?

simoneb (approved):

PR #771, opened 2022-01-20
fix(typing): add undefined to FetchData Promise return type

### LINKED ISSUE #772: The typing of FetchData is incorrect with regards to its return type
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: 5.5.1

### Description

useClientRequest's fetchData function returns an empty (undefined) promise when the hook is unmounted, but this case is not reflected in the typing. This also affects the fetcher functions returned by useMutation and useManualQuery.

### How to reproduce

The behavior is evident in useClientRequest.js:136 and also tested for in useClientRequest.test.js:476 `it('returns undefined instantly if not mounted'`

I've created pull request #771 which adds `undefined` to the Promise return type of the FetchData type.

Another solution would be to return a `{ loading: false, cacheHit: false }` object (which would satisfy `UseClientRequestResult<ResponseData, TGraphQLError>`)  instead of the empty `Promise.resolve()` in `useClientRequest` - this would not be a breaking change for Typescript users, and I don't imagine many depends on `undefined` being returned in this edge case. I would be happy to make another PR with this solution if the maintainers think it would be more appropriate.

csorfab:
The fetchData function of useClientRequest returns an empty Promise if the hook is unmounted. Typing has been updated to reflect this behavior

### What does this PR do?

It fixes a typing error that affects useClientRequest's fetchData function, and as a result, useManualQuery and useMutation too. useClientRequest's fetchData function returns an empty (undefined) Promise if called after the component has unmounted, but this is not currently reflected in its typing. Admittedly, this is an edge case behavior, but it's caused my team hard to reproduce bugs, and since the behavior is codified in useClientRequest.test.js:476 ("returns undefined instantly if not mounted"), it seems appropriate that the typing reflect it.

This would be a breaking change for many projects, but I believe that Typescript's usefulness lies exactly in preventing hard-to-reproduce bugs like those caused by this behavior, so I think it's a small price to pay for correctness.

### Related issues
 Closes #772 

### Checklist

### DIFF
--- a/packages/graphql-hooks/index.d.ts
+++ b/packages/graphql-hooks/index.d.ts
@@ -238,7 +238,7 @@ interface UseSubscriptionOperation<Variables extends object = object>
 
 type FetchData<ResponseData, Variables = object, TGraphQLError = object> = (
   options?: UseClientRequestOptions<ResponseData, Variables>
-) => Promise<UseClientRequestResult<ResponseData, TGraphQLError>>
+) => Promise<UseClientRequestResult<ResponseData, TGraphQLError> | undefined>
 
 interface CacheKeyObject {
   operation: Operation
--- a/packages/graphql-hooks/test-d/hooks.test-d.ts
+++ b/packages/graphql-hooks/test-d/hooks.test-d.ts
@@ -1,4 +1,4 @@
-import { expectType } from 'tsd'
+import { expectType, expectError } from 'tsd'
 import {
   FetchData,
   ResetFunction,
@@ -35,6 +35,10 @@ type UseClientRequestReturn = [
 ]
 expectType<UseClientRequestReturn>(useClientRequest(query))
 expectType<UseClientRequestReturn>(useClientRequest(query, clientRequestOptions))
+
+const [fetchData] = useClientRequest(query);
+expectType<Promise<UseClientRequestResult<any, object> | undefined>>(fetchData());
+expectError<Promise<UseClientRequestResult<any, object>>>(fetchData());
 
 const useQueryOptions: UseQueryOptions = {
   ssr: false,


simoneb:
can you please create an issue and use this PR to propose a solution for it?

csorfab:
Hi @simoneb, I've created issue #772 and linked to this PR in it.

csorfab:
Another solution would be to return a `{ loading: false, cacheHit: false }` object instead of the empty `Promise.resolve()` in useClientRequest - this would not be a breaking change for Typescript users, and I don't imagine many depends on `undefined` being returned in this edge case. I got discouraged from this route since the tests explicitly test for behavior, but I think it might be a better solution than what the current PR achieves? @simoneb what do you think about this?

simoneb:
Can you please cover this change with a test by changing or adding to the code in https://github.com/nearform/graphql-hooks/tree/master/packages/graphql-hooks/test-d?

simoneb:
@csorfab ping :)

csorfab:
@simoneb sorry for the delay! I checked the typing tests with `npm run tsd` and it ran fine, so I added to assertions that covers the change in the PR.

Have you considered the other option, which would be returning a "dummy" object (`{ loading: false, cacheHit: false }`) instead of `undefined`? I'm beginning to see it as a better option, because it feels like more consistent behavior - maybe an Error could be included, describing that the request was aborted because the hook unmounted? Certainly feels better than just returning `undefined` without any explanation

tiagoheliob:
@csorfab Could you please draft an implementation using your suggestion of returning a dummy object? it's certainly more suitable than an undefined return for sure as you mentioned

csorfab:
@tiagoheliob Hi! I've created a new PR with the implementation of my other suggestion: #777

tiagoheliob:
@csorfab Thanks a lot, I'll take a look

csorfab:
I'm closing this as #777 has been merged instead

PR #777, opened 2022-01-27
feat(fetchData): no undefined when unmounted

### LINKED ISSUE #772: The typing of FetchData is incorrect with regards to its return type

csorfab:
useClientRequests's fetchData no longer returns undefined when the hook is unmounted,
instead returns a result object with a descriptive error

### What does this PR do?

It resolves #772 by returning a result object conforming to the typing and normal behavior of useClientRequest's fetchData

### Related issues

#772 

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/useClientRequest.js
+++ b/packages/graphql-hooks/src/useClientRequest.js
@@ -133,7 +133,6 @@ function useClientRequest(query, initialOpts = {}) {
   // arguments to fetchData override the useClientRequest arguments
   const fetchData = useDeepCompareCallback(
     newOpts => {
-      if (!isMounted.current) return Promise.resolve()
       const revisedOpts = {
         ...initialOpts,
         ...newOpts
@@ -143,6 +142,18 @@ function useClientRequest(query, initialOpts = {}) {
         ...operation,
         variables: revisedOpts.variables,
         operationName: revisedOpts.operationName
+      }
+
+      if (!isMounted.current) {
+        return Promise.resolve({
+          error: {
+            fetchError: new Error(
+              'fetchData should not be called after hook unmounted'
+            )
+          },
+          loading: false,
+          cacheHit: false
+        })
       }
 
       const revisedCacheKey = client.getCacheKey(revisedOperation, revisedOpts)
--- a/packages/graphql-hooks/test/unit/useClientRequest.test.js
+++ b/packages/graphql-hooks/test/unit/useClientRequest.test.js
@@ -473,7 +473,7 @@ describe('useClientRequest', () => {
       expect(state).toEqual({ cacheHit: false, loading: true })
     })
 
-    it('returns undefined instantly if not mounted', async () => {
+    it('returns result with error instantly if not mounted', async () => {
       let fetchData, state
       const { unmount } = renderHook(
         () =>
@@ -487,8 +487,16 @@ describe('useClientRequest', () => {
       )
 
       unmount()
-      const result = await act(fetchData)
-      expect(result).toBe(undefined)
+
+      let result
+      await act(async () => {
+        result = await fetchData()
+      })
+
+      expect(result.data).toBe(undefined)
+      expect(result.loading).toBe(false)
+      expect(result.cacheHit).toBe(false)
+      expect(result.error.fetchError).toBeInstanceOf(Error)
       expect(mockClient.request).not.toHaveBeenCalled()
       expect(state).toEqual({ cacheHit: false, loading: true })
     })


packages/graphql-hooks/src/useClientRequest.js
```diff
@@ -145,6 +144,18 @@ function useClientRequest(query, initialOpts = {}) {
         operationName: revisedOpts.operationName
       }
 
+      if (!isMounted.current) {
```
tiagoheliob, line 147: Doesn't it feel weird to use `Promise.resolve`? instead of `Promise.reject` to return an error?

tiagoheliob:
@csorfab just let a comment about returning `Promise.reject` instead of `Promise.resolve`

packages/graphql-hooks/src/useClientRequest.js
csorfab: @tiagoheliob it does feel a bit weird, but I was trying to conform to existing behavior as much as possible while providing a useful error message - the only other cases where this hook throws errors/rejects are input type checks, so I would find it a bit unexpected to get a throw here. If it is decided that it should reject instead, I'd be happy to update my PR, maybe @simoneb can chime in on this one?

simoneb:
@csorfab thanks for all the work you've done here. My main concern with this change is that it's a breaking change and because I don't know the motivations that led to the current implementation I wouldn't feel comfortable changing it right now. 

If you agree with these considerations, can we go back to your first PR and only get the typing change in?

csorfab:
@simoneb Well, the thing is, the other PR would be a **major** breaking change for Typescript users, for example it would be no longer possible to destructure the result of an awaited fetchData because you can't destructure undefined. This PR is in line with the typing, and while I obviously don't know the original intent behind the empty resolve(), it definitely feels more like a quick hack, than intended behavior. Calling fetchData after it unmounted is an edge case, and I have a hard time imagining that anyone's code would depend on fetchData returning undefined in that very specific edge case - you always need to check for result.error under normal circumstances, and the original behavior causes bugs with that, that's what happened with me and what motivated this PR. If anything, I think this would fix unnoticed bugs in people's code rather than break anything.

csorfab:
@simoneb I dug up the PR that originally introduced this behavior: #96 (related issue: #54)

It seems to me that no particular thought was given to what should be returned in this case, it was just a quick bug fix

packages/graphql-hooks/src/useClientRequest.js
tiagoheliob: That's true, I agree!

tiagoheliob:
@csorfab https://github.com/nearform/graphql-hooks/pull/96 seems like a quick fix like you mentioned, what Simone meant is that people are currently doing something like the example below would have these changes to break their code

current 
```
const  something = await fetchData(...)

if (!something) { 
 // do something 
 return;
}
// do something else
```

with the introduced changes 
```
const  something = await fetchData(...)

if (something.error...) { 
 // do something 
 return;
}
// do something else
```

csorfab:
@tiagoheliob I understand, but are you sure that people actually do this? Because this behavior is not mentioned anywhere in the docs, nor the typings, and in 99.99% of cases the `if(!something)` banch will never get called, so it's impossible to find out about it during normal usage (where you follow the something.error/something.data pattern based on the docs).

The only way to find out about this behavior is by looking at the source code after unexpected, hard to reproduce errors (which is how I came about it)

tiagoheliob:
@csorfab good point, I've made some tests and it's working fine, I'll approve it

tiagoheliob (approved):
LGTM

simoneb (approved):

csorfab:
Thank you all for the cooperation!

simoneb:
Thank you @csorfab for your work and the thought you put into this! This change is now released in the latest version of the package.

Issue #759, opened 2022-01-04
RFC: Support for Automatic Persisted Queries (APQ)


zanettin:
### Description

We offer currently a persisting method thanks to @salmanm's [PR](https://github.com/nearform/graphql-hooks/pull/493). with this method we create the hashes on build time and they need to be available on the backend as well. with the automatic approach the client would hash the query on request and sends it to the server. if the server already has a cached response for this hash, it would return it. otherwise he'll send an error code after which the client has to resend the request using the query and the hash. illustrated here:
![Bildschirmfoto 2022-01-04 um 10 48 55](https://user-images.githubusercontent.com/3241476/148040723-ede29fb0-be40-4add-b4fb-bbdd901ff356.png)

### Server Implementations with APG support
- [Apollo](https://www.apollographql.com/docs/apollo-server/performance/apq/)
- [Dgraphq](https://dgraph.io/docs/graphql/queries/persistent-queries/)
- (more?)

### Suggested implementation
- add a new option on request level to enable it
- on request:
  - hash the query using sha256
  - enforce `GET` request
  - change request format to smth like. `curl -g 'http://localhost:4000/graphql?extensions={"persistedQuery":{"version":1,"sha256Hash":"ecf4edb46db40b5132295c0291d62fb65d6759a9eedfa4d5d612dd5ec54a6b38"}}'`
  - on success: nothing has to be changed
  - on error:
    - if error type `PERSISTED_QUERY_NOT_FOUND`: resend query using `POST` together with the hash
    - other errors are handled as they're currently handled

### Questions
- do we want to offer these option on the client?
- how would we like to enable it to not conflict with the existing `persisted` option

simoneb:
Hi @zanettin, before getting into the details of this request there's one thing I would like to clarify. You mention 

> if the server already has a cached response for this hash, it would return it

Are we talking about persisted queries as a caching mechanism or am I not understanding this right?

zanettin:
Hi @simoneb

Sorry that I was not clear about that point. It's not caching related but eliminates the step of persisting the queries at build time. As visualized on the sequence diagram above the client sends always just an sha256 hash to the graphql server. if the server knows the hash, he translates it to the real query and executes it. if the server is not aware of that hash since he never had to answer a request with that hash, he returns an error to the client. the client then has to send the query alongside with the sha256 hash. the server will execute the query and if it was successful he'll stores the hash/query map on any (shared) cache.

the cache could look like this:
```
{ 
[...]

PR #791, opened 2022-03-09
Add middleware support

### LINKED ISSUE #769: Add middleware support
### Discussed in https://github.com/nearform/graphql-hooks/discussions/761

This feature is needed to add features like Automatic Persisted Queries (#759), or for any other advanced usage.

In apollo they enable such use cases with _links_:

https://www.apollographql.com/docs/react/api/link/introduction/
https://github.com/apollographql/apollo-client/tree/main/src/link

There's also an interesting approach from `urql`, called _exchanges_:

https://formidable.com/open-source/urql/docs/architecture/#the-client-and-exchanges
https://formidable.com/open-source/urql/docs/api/core/#exchanges

It seems they claim that's more flexible [here](https://formidable.com/open-source/urql/docs/comparison/) (see the cache exchange point compared with Apollo).

vaclav-zeman:
### What does this PR do?

- Adds a new GraphQLClient config option for middleware - functions that are run in order before executing a request and that can transform the request before it's sent
- Ability to access and transform response with `addResponseHook`
- Ability to break the control flow with `resolve/reject` - afaik needed for APQ midd
- This is far from perfect and other libraries (like urql) have more sophisticated solutions using streams, that seem to be better suited. However, the advantage of this solution is that it keeps the minimal size of this lib 
- Some example middlewares (very naive implementations) are added to make sure they could be used with this approach 

**Very open to feedback**, I think there's quite a room for improvement but not sure what would be the best for this particular library

### Related issues

Close #769
Relates to #759 

### Checklist

- [x] I have added or updated any relevant documentation **Will add when the functionality is finalized**

### DIFF
--- a/packages/graphql-hooks/src/GraphQLClient.js
+++ b/packages/graphql-hooks/src/GraphQLClient.js
@@ -2,6 +2,7 @@ import EventEmitter from 'events'
 import canUseDOM from './canUseDOM'
 import { extractFiles } from 'extract-files'
 import isExtractableFileEnhanced from './isExtractableFileEnhanced'
+import Middleware from './Middleware'
 
 class GraphQLClient {
   constructor(config = {}) {
@@ -56,6 +57,17 @@ class GraphQLClient {
     this.logErrors = config.logErrors !== undefined ? config.logErrors : true
     this.onError = config.onError
     this.useGETForQueries = config.useGETForQueries === true
+    this.middleware = new Middleware()
+    const middlewares = [
+      (opts, next) => {
+        console.log('LOGGER', opts)
+        next()
+      }
+    ]
+      .concat(config.middleware)
+      .filter(Boolean)
+    middlewares.forEach(mid => this.middleware.use(mid))
+
     this.mutationsEmitter = new EventEmitter()
   }
 
@@ -201,15 +213,20 @@ class GraphQLClient {
     return fetchOptions
   }
 
-  request(operation, options = {}) {
-    if (this.fullWsTransport) {
-      return this.requestViaWS(operation)
-    }
+  request(rawOperation, options = {}) {
+    return new Promise(resolve =>
+      this.middleware.go({ operation: rawOperation }, ({ operation }) => {
+        console.log(operation)
+        if (this.fullWsTransport) {
+          return resolve(this.requestViaWS(operation))
+        }
 
-    if (this.url) {
-      return this.requestViaHttp(operation, options)
-    }
-    throw new Error('GraphQLClient: config.url is required')
+        if (this.url) {
+          return resolve(this.requestViaHttp(operation, options))
+        }
+        throw new Error('GraphQLClient: config.url is required')
+      })
+    )
   }
 
   requestViaHttp(operation, options) {
--- a/packages/graphql-hooks/src/GraphQLClient.js
+++ b/packages/graphql-hooks/src/GraphQLClient.js
@@ -261,13 +261,12 @@ class GraphQLClient {
       url = url + '?' + paramsQueryString
 
       if (operation.hash) {
-        url += `&extensions=${JSON.stringify(
-          {
+        const extensions = encodeURIComponent(
+          JSON.stringify({
             persistedQuery: { version: 1, sha256Hash: operation.hash }
-          },
-          null,
-          0
-        )}`
+          })
+        )
+        url += `&extensions=${extensions}`
       }
     }
 
--- a/packages/graphql-hooks/src/GraphQLClient.js
+++ b/packages/graphql-hooks/src/GraphQLClient.js
@@ -210,16 +210,19 @@ class GraphQLClient {
 
   request(rawOperation, options = {}) {
     return new Promise((resolve, reject) =>
-      this.middleware.go({ operation: rawOperation }, ({ operation }) => {
-        if (this.fullWsTransport) {
-          return resolve(this.requestViaWS(operation))
-        }
+      this.middleware.go(
+        { operation: rawOperation, resolve, reject },
+        ({ operation }) => {
+          if (this.fullWsTransport) {
+            return resolve(this.requestViaWS(operation))
+          }
 
-        if (this.url) {
-          return resolve(this.requestViaHttp(operation, options))
+          if (this.url) {
+            return resolve(this.requestViaHttp(operation, options))
+          }
+          reject(new Error('GraphQLClient: config.url is required'))
         }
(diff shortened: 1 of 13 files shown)


packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -230,7 +257,18 @@ class GraphQLClient {
           return `${k}=${encodeURIComponent(v)}`
         })
         .join('&')
+
       url = url + '?' + paramsQueryString
+
+      if (operation.hash) {
+        url += `&extensions=${JSON.stringify(
```
galvez, line 264: Might want to `encodeURIComponent()` here for safety.

packages/graphql-hooks/src/middlewares/apqMiddleware.js
```diff
@@ -0,0 +1,39 @@
...
+    // Add SHA256 of a query
+    operation = { ...operation, hash: await sha256(operation.query) }
+
+    // Try to send just the hash
+    const res = await client.requestViaHttp(operation, {
+      fetchOptionsOverrides: { method: 'GET' },
+      hashOnly: true
+    })
+
+    if (!res.error) resolve(res)
```
galvez, line 25: I'd avoid inline ifs for consistency (always seeing curly braces for JS flow constructs) and readability. Just a suggestion.

simoneb (commented):
One question before reviewing, did we inspire from any specific source for the design of this middleware implementation?

vaclav-zeman:
Hey @simoneb , good q - the generic middleware pattern is from this small gist https://gist.github.com/darrenscerri/5c3b3dcbe4d370435cfa

simoneb:
@vaclav-zeman to aid with the review of this change, which we're doing for the primary purpose of supporting the implementation of #759, would it be worth sharing some pseudocode showing how that feature can be implemented with this middleware support? That may drive further iterations on the current implementation.

vaclav-zeman:
@simoneb The pseudocode for APQ is already here - https://github.com/nearform/graphql-hooks/blob/63e46110cc29260292894166a6efc2c773ae1f25/packages/graphql-hooks/src/middlewares/apqMiddleware.js  Unless I totally misunderstood the APQ solution completely :)

packages/graphql-hooks/src/middlewares/apqMiddleware.js
```diff
@@ -0,0 +1,42 @@
...
+
+    // Try to send just the hash
+    const res = await client.requestViaHttp(operation, {
+      fetchOptionsOverrides: { method: 'GET' },
+      hashOnly: true
+    })
+
+    // Data fetched successfully -> resolve early
+    if (!res.error) {
+      resolve(res)
```
simoneb, line 27: you should return here otherwise the execution will continue

floridemai:
@vaclav-zeman The readme says lack of middleware to be one of the cons of the library. That will not be the case anymore. I would make changes to the readme files and maybe add something around configuring middleware.

vaclav-zeman:
@floridemai yep, I was planning to update the readme after any potential updates to this PR

packages/graphql-hooks/src/Middleware.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    for (const fn of fns) {
+      if (typeof fn !== 'function') {
+        throw new Error(
+          'GraphQLClient Middleware: middleware has to be of type `function`'
+        )
+      }
+
+      this.run = (stack => (opts, next) => {
```
floridemai, line 28: Would be nice to avoid anonymous functions in order to debug (maybe even test it tbh) easier.
```
      at apply (packages/graphql-hooks/test/unit/Middleware.test.js:6:11)
      at apply (packages/graphql-hooks/src/Middleware.js:31:16)
      at stack (packages/graphql-hooks/src/Middleware.js:57:10)
      at stack (packages/graphql-hooks/src/Middleware.js:30:11)
      at Middleware.run (packages/graphql-hooks/src/Middleware.js:30:11)
      at Object.<anonymous> (packages/graphql-hooks/test/unit/Middleware.test.js:35:7)
```

vaclav-zeman: Good idea. Will create a new issue for this as the main priority is now to support APQ middleware

packages/graphql-hooks/src/middlewares/apqMiddleware.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * AutomaticPersistedQueryMiddleware - must be last in the middleware list
+ * @param {function} makeRequest
+ * @returns Promise<object>
+ */
+const APQMiddleware = async ({ operation, client, resolve, reject }, next) => {
+  try {
+    // Add SHA256 of a query
+    operation = { ...operation, hash: await sha256(operation.query) }
```
floridemai, line 17: If we want something solid (and proved to be working) we can use something like this:
https://github.com/rickbergfalk/postgrator/blob/master/lib/utils.js#L39
It is true that literally any change will make an impact, but that might be ok?

floridemai (approved):
LGTM.
Enough tests to prove this usable right away.

packages/graphql-hooks/src/middlewares/apqMiddleware.js
vaclav-zeman: Two concerns: 1) I assumed APQ needs sha256, 2) need to check if the crypto lib is present in the browser. I'll take a look in the next PR, thanks 👍

packages/graphql-hooks/src/middlewares/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+See [examples folder](examples/)
+
+### Update response post-request
+
+```js
+const camelCaseMiddleware = ({ operation, addResponseHook }, next) => {
+  addResponseHook(response => {
+    // Need to return
+    return toCamelCaseDeep(response)
```
simoneb, line 38: what would this do? doesn't feel like a typical operation somebody would want to do. maybe a more realistic example?

simoneb (approved):

packages/graphql-hooks/src/middlewares/README.md
vaclav-zeman: Right, I had a use case for something similar with rest api that returned snake case vars and later were camel cased so got inspiration in there. Will think of something more realistic 👍

simoneb: thanks. not a big deal, let's just make sure it's clear what we're trying to do

PR #794, opened 2022-03-14
Add APQ Middleware

### LINKED ISSUE #759: RFC: Support for Automatic Persisted Queries (APQ)

vaclav-zeman:
### What does this PR do?

- Add APQ Middleware with proper hashing function 
- The skeleton was already here so this was mostly code cleanup and update 
image
- TODO: should properly test it with a standalone GraphQL API with APQ support

### Related issues

Close #759
Blocked by #791

### Checklist


vaclav-zeman:
For the reviewer: I didn't manage to test this against a standalone Apollo server before leaving the bench, it would be very welcome if you could try it. Thank you 🙌

packages/graphql-hooks/package.json
```diff
@@ -35,6 +35,7 @@
     "react": "^17.0.0"
   },
   "dependencies": {
+    "@aws-crypto/sha256-browser": "^2.0.1",
```
radomird, line 38: Is there a specific reason why you used the browser compatible version here? I see that there is a "regular" js version [sha256-js](https://github.com/aws/aws-sdk-js-crypto-helpers/tree/master/packages/sha256-js) as well as [sha256-universal](https://github.com/aws/aws-sdk-js-crypto-helpers/tree/master/packages/sha256-universal).

vaclav-zeman: The only reason is that the browser version has a fallback for when `window.crypto.subtle` is not present. I didn't spot the universal though - that would work too but it's a little bigger lib

radomird:
Hi @vaclav-zeman I tested with a standalone Apollo server and it seems it's not compatible - mainly this part of the code:
```js
if (JSON.parse(error.fetchError).type === 'PERSISTED_QUERY_NOT_FOUND') {
      next()
 } else {
      throw error
 }
```

Apollo server returns `error.graphQLErrors` array with all the errors so then we need to read this property, to get the error code: `error.graphQLErrors[0].extensions.code` ([0] is just the first error in the array, I used this as an example).
You can find more info in the [Apollo server docs](https://www.apollographql.com/docs/apollo-server/data/errors/).

I will update this PR to handle this case as well.

simoneb:
@radomird this behavior is a little suspicious. First I don't find confirmation of what you wrote in the Apollo docs. They say that Apollo returns an `errors` array, not `graphQLErrors`. So I wonder if we're doing any transformation in here

simoneb:
> * doesn't handle errors in the `graphQLErrors` which are populated here: https://github.com/nearform/graphql-hooks/blob/52f3fa02e16b6446b8f5163afb2fc3de643ba484/packages/graphql-hooks/src/GraphQLClient.js#L290
> * the error is a stringified object:
>   ```js
>    JSON.parse(error.fetchError).type
>   ```

can we fix this? do we have to change the core middleware implementation or can we change the APQ middleware alone?

packages/graphql-hooks/package.json
simoneb, line 38: does this work on the server? this package should also work in SSR scenarios

packages/graphql-hooks/rollup.config.js
```diff
@@ -1,3 +1,14 @@
...
-export default generateRollupConfig('GraphQLHooks')
+export default generateRollupConfig('GraphQLHooks').concat({
+  input: 'src/middlewares/apqMiddleware.js',
+  output: {
+    file: `lib/middlewares/apqMiddleware.js`,
+    format: 'cjs',
+    indent: false
+  },
+  plugins: [babel(), sizeSnapshot()]
+})
```
simoneb, line 14: not sure what this is, mind clarifying?

packages/graphql-hooks/src/middlewares/README.md
```diff
@@ -25,6 +25,22 @@ Middleware is a function with 2 parameters:
   - `resolve`, `reject` - advanced usage only, these would be used only when we want to resolve/reject the request `Promise` early without doing the usual fetch (see [example](examples/cacheMiddleware.js))
 - `next` function - calls the next middleware function in line. Generally it should always be called, unless we want to change the control flow
 
+## Automatic Persisted Queries (APQ) middleware
+
+This library also includes an Automatic Persisted Queries middleware that can improve network performance by sending smaller requests. The implementation should match the Apollo back-end functionality that is [further described here](https://www.apollographql.com/docs/apollo-server/v2/performance/apq/).
+
+Usage,
```
simoneb, line 32: "Usage," doesn't look like should be formatted this way

packages/graphql-hooks/src/middlewares/README.md
```diff
@@ -25,6 +25,22 @@ Middleware is a function with 2 parameters:
   - `resolve`, `reject` - advanced usage only, these would be used only when we want to resolve/reject the request `Promise` early without doing the usual fetch (see [example](examples/cacheMiddleware.js))
 - `next` function - calls the next middleware function in line. Generally it should always be called, unless we want to change the control flow
 
+## Automatic Persisted Queries (APQ) middleware
```
simoneb, line 28: okay I'm not sure about this decision. why are we not including this automatically?

simoneb (commented):
A few comments

packages/graphql-hooks/rollup.config.js
vaclav-zeman: this makes sure that apqMiddleware is build separately from the main bundle (so no need to have it in the bundle if not used)

packages/graphql-hooks/src/middlewares/README.md
vaclav-zeman: I thought it's better to give users a choice but if you think it's a good default, then np => can be bundled inside the main lib and instantiated + passed in [here](https://github.com/nearform/graphql-hooks/blob/8451502568468629b038ba054727a7d037dbd962/packages/graphql-hooks/src/GraphQLClient.js#L61) 👍 

My main concern was that the lib size would no longer be that small which seems to be one of the highlights of this lib (looking at readme)

packages/graphql-hooks/package.json
vaclav-zeman: Using `@aws-crypto/sha256-universal` would be a better choice as @radomird pointed out previously

packages/graphql-hooks/src/middlewares/apqMiddleware.js
```diff
@@ -1,9 +1,14 @@
...
+
+export async function sha256(query) {
+  const hash = new Sha256()
+  hash.update(query, 'utf8')
+  const hashBuffer = await hash.digest()
+  // convert buffer to byte array
+  const hashArray = Array.from(hashBuffer)
+  // convert bytes to hex string
+  const hashHex = hashArray.map(b => b.toString(16).padStart(2, '0')).join('')
+  return hashHex
```
wilkmaia, line 11: This might be a bit cleaner, but it's not that big of a deal IMO:

```suggestion
  const hashUint8Array = await hash.digest()

  const hashBuffer = Buffer.from(hashUint8Array)
  return hashBuffer.toString('hex')
```

The `Buffer` approach should also perform better due to Node's internal optimizations. A quick benchmark I've ran showed it's a little bit over 50% faster, with different queries (to avoid any caching) with length `100`.

simoneb:
@radomird @wilkmaia what's the status of this?

radomird:
> @radomird @wilkmaia what's the status of this?

@simoneb Vaclav responded to your comments - since then I didn't check this PR. Can you please check his reply to your comments. The comments don't have a direct correlation to my work on this PR but I'm happy to address any changes needed.

packages/graphql-hooks/src/middlewares/README.md
kurtmilam: Should be `Usage:` to be consistent with the rest of the doc.

packages/graphql-hooks/src/middlewares/apqMiddleware.js
kurtmilam: Agreed on this - let's use built-ins where possible, unless there's a compelling reason not to.

packages/graphql-hooks/package.json
kurtmilam: Agreed - `@aws-crypto/sha256-universal` seems to be the appropriate choice here.

packages/graphql-hooks/src/middlewares/README.md
kurtmilam: I don't have full context on this library, but I tend to side with vaclav on this one.

kurtmilam (changes requested):
I'll address the changes I commented on.

wilkmaia (approved):

wilkmaia (approved):

radomird (approved):
lgtm

README.md
```diff
@@ -972,162 +971,6 @@ function MyComponent() {
 
 - `called`
 
-## Testing and mocking
```
simoneb, line 975: why is this section gone?

kurtmilam: mea culpa - certainly wasn't intended - I may have missed updating locally after merging `master` into the feature branch. I checked the other later commits on this, and nothing else snuck into those.

kurtmilam: I've restored the missing content.

simoneb (approved):

Issue #796, opened 2022-03-18
Migrate the codebase to typescript


gdorsi:
### Description

When adding/changing features is easy to forgot to update the typescript types. (see https://github.com/nearform/graphql-hooks/issues/795)

I think migrating to Typescript would make it easier to contribute to the project because the types would no longer be separated from the source code.

PR #797, opened 2022-03-21
feat: local mock client for graphql-hooks

### LINKED ISSUE #746: Add a Testing section in the docs

liana-pigeot:
### What does this PR do?

This is a mock client that doesn't connect to any servers and instead has a list of queries and their premade replies for ease of use during testing, or as a mock during front-end development.

Based on PR #751 by Guido

The new `LocalGraphQLClient` extends `GraphQLClient` but replaces the `request` method with a simple way to respond to pre-defined queries.

 Queries are given to the local client in the form of an object of functions:

 ```
    const localQueries = {
      [allPostsQuery]: () => ({
        allPosts: [
          {
            id: 1,
            title: 'Test',
            url: 'https://example.com'
          }
        ]
      }),
      [createPostMutation]: () => ({ createPost: { id: 1 } }),
    }
    const client = new LocalGraphQLClient({ localQueries })
  ```

PR in progress: Tests are passing but I'd like feedback on the solution, and still need to add documentation and unit tests. I also need to update the type definition for the package

### Related issues

closes #746 

### Checklist


packages/graphql-hooks/index.d.ts
```diff
@@ -122,7 +123,8 @@ export const ClientContext: React.Context<GraphQLClient>
 type ResetFunction = (desiredState?: object) => void
 
 interface ClientOptions {
-  url: string
+  url?: string
+  local?: LocalQueries
```
simoneb, line 127: this is no longer needed if we're extending the client, right?

packages/graphql-hooks/index.d.ts
```diff
@@ -7,6 +7,7 @@ import { SubscriptionClient } from 'subscriptions-transport-ws'
 export class GraphQLClient {
   constructor(options: ClientOptions)
 
+  local: LocalQueries
```
simoneb, line 10: this is no longer needed if we're extending the client, right?

packages/graphql-hooks/src/GraphQLClient.js
```diff
@@ -15,6 +15,29 @@ class GraphQLClient {
...
+    this.logErrors = config.logErrors !== undefined ? config.logErrors : true
+    this.onError = config.onError
+    this.useGETForQueries = config.useGETForQueries === true
+    this.middleware = new Middleware(config.middleware || [])
+
+    this.mutationsEmitter = new EventEmitter()
+  }
+
+  /** Checks that the given config has the correct required options */
+  verifyConfig(config) {
```
simoneb, line 40: not sure I understand what happened here. has the code only been moved around? please clarify because it's not obvious by just reading the code

packages/graphql-hooks/src/LocalGraphQLClient.js
```diff
@@ -0,0 +1,54 @@
...
+            url: 'https://example.com'
+          }
+        ]
+      }),
+      [createPostMutation]: () => ({ createPost: { id: 1 } }),
+    }
+    const client = new LocalGraphQLClient({ localQueries })
+  ```
+ */
+class LocalGraphQLClient extends GraphQLClient {
```
simoneb, line 23: I wonder, do we really need to extend the client? Why do we need to inherit from it? Is there anything that the subclass needs from the base class?

packages/graphql-hooks/src/LocalGraphQLClient.js
```diff
@@ -0,0 +1,54 @@
...
+  constructor(config = {}) {
+    super(config)
+    this.localQueries = config.localQueries
```
simoneb, line 26: one reason why I'm not sure we need to extend the class is that I believe that the configuration options of the real client largely (or entirely even) don't apply to the local one. In fact I'd go as far as having the local client take a single constructor argument which is the local queries. am I missing something?

packages/graphql-hooks/src/LocalGraphQLClient.js
```diff
@@ -0,0 +1,54 @@
...
+          operation.variables,
+          operation.operationName
+        )
+      ).then(data => ({ data }))
+    } else {
+      throw new Error(
+        `LocalGraphQLClient: no query match for: ${operation.query}`
+      )
+    }
+  }
```
simoneb, line 51: invert condition to return early. it's in general a good practice and will avoid a level of nesting of the code here (aka cyclomatic complexity)

packages/graphql-hooks/src/GraphQLClient.js
liana-pigeot: Yes, just moved around so that the local version of the client can override this function to remove all those checks

packages/graphql-hooks/index.d.ts
liana-pigeot: I still need to update those definitions with the new scheme, I think this bit was originally added by Guido with the previous implementation of the local client

packages/graphql-hooks/src/LocalGraphQLClient.js
liana-pigeot: It's mostly to make it easily conform to the same interface and accept the same options (even if most will end up unused in the case of the local client). It could probably work even if we don't extend I suppose, but this way the relationship between the two is pretty clear

packages/graphql-hooks/src/LocalGraphQLClient.js
liana-pigeot: Pretty much, but I think the difference is in how you might use it in testing/development code. Even if this does nothing, it will still have methods available to be called (if you look in the parent class, things like `setHeader`, or cache functions, or even the logging). This way we can have it behave as close as possible to the original one and it should be possible to run existing code by just replacing the "remote" client with a "local" one.

If we don't inherit, the local client might need to manually implement or mock more functions, which I think was what people were trying to avoid in the original PR

liana-pigeot: Another option I thought about was to have a `RemoteGraphQLConnection` and `LocalGraphQLConnection`, and the client takes a `Connection` parameter, which would mean we're always using the real client and we're just replacing the object that takes care of sending requests.

I thought it was probably a bit overkill though so kept it simple for now

simoneb: In this case I have a feeling that what we're looking at is implementing an interface then, rather than inheriting a class, because the only thing that the local client needs from the real client is the interface and no implementation. 

Nevertheless, I suppose that doing it in that way is a bigger change, at least as far as types are concerned, so I'm good with inheritance as long as we agree that it's a compromise

simoneb: > I thought it was probably a bit overkill though so kept it simple for now

yes, not suggesting to do something overkill

liana-pigeot: also this repo isn't in typescript, so there's no compiler to verify that the local client actually conforms to the interface of the real client easily

simoneb: but we have typings, don't we?

packages/graphql-hooks/index.d.ts
gdorsi: It's good to have this comment as a reminder to update the types.

It's something easy to forget :)

packages/graphql-hooks/src/LocalGraphQLClient.js
gdorsi: Unfortunately the internal hooks like `useClientRequest` and `useQuery` [depends](https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks/src/useClientRequest.js#L159) on [many](https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks/src/useClientRequest.js#L201) [methods](https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks/src/useQuery.js#L82) from `GraphQLClient`.

By extending we are sure that we don't break any of these dependencies.

Would be nice to be able to do something like [urql](https://formidable.com/open-source/urql/docs/advanced/testing/#handling-multiple-hooks) where to create a client is enough to provide an object with the request methods.

Since the access to `GraphQLClient` and all it's methods is public, to get that we would have to introduce some breaking changes.

examples/create-react-app/test/Posts.test.js
```diff
@@ -0,0 +1,29 @@
...
+    render(<Posts />, {
+      localQueries
+    })
+
+    expect(
+      await screen.findByRole('link', {
+        name: /Test/i
+      })
+    ).toBeTruthy()
+  })
```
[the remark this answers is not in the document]
liana-pigeot: Good idea, I still need to add documentation for testing in the readme, so I can add examples there when I get to it

packages/graphql-hooks/src/LocalGraphQLClient.js
```diff
@@ -0,0 +1,54 @@
...
+
+  request(operation) {
+    if (this.localQueries[operation.query]) {
+      return Promise.resolve(
+        this.localQueries[operation.query](
+          operation.variables,
+          operation.operationName
+        )
+      ).then(data => ({ data }))
```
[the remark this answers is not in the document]
liana-pigeot: Good point, I'll look at adding that

packages/graphql-hooks/src/LocalGraphQLClient.js
simoneb: no you're right!

packages/graphql-hooks/src/LocalGraphQLClient.js
simoneb: That's why I mentioned it's just an API issue, not a functionality issue. The local client uses little to nothing of the real client, so an interface would do. On the other hand, I believe that turning it into an interface will be a breaking change, hence why I'm ok with inheritance

packages/graphql-hooks/src/LocalGraphQLClient.js
simoneb: @liana-pigeot ^ bump

packages/graphql-hooks/index.d.ts
simoneb: let's add it as an issue, but clearly not to be addressed in this same PR

simoneb:
Ok @gdorsi @liana-pigeot I'm happy with the approach taken here, with the only exception (which I don't feel too strongly about) that inheriting from a class just for the sake of the interface is wrong. Let's please spend some time about thinking whether we have other options. I'm not sure if you were around when OOP was a thing, but that's really one of the basic rules, or in this case, antipatterns. 

Apart from this, the underlying approach of creating something which can be swapped and allows to mock the queries is perfect!

packages/graphql-hooks/src/LocalGraphQLClient.js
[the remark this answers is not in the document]
liana-pigeot: As I was implementing this, I realised the normal GraphQL client never actually throws errors, it just returns an `error` object (which is not itself an error), as done by this code:

```
  generateResult({ fetchError, httpError, graphQLErrors, data }) {
    const errorFound = !!(
      (graphQLErrors && graphQLErrors.length > 0) ||
      fetchError ||
      httpError
    )
    return !errorFound
      ? { data }
      : { data, error: { fetchError, httpError, graphQLErrors } }
  }
```

Which  means, the local mock should not throw either but instead should return the "error" as part of the object the query returns.

But to keep the convenience of being able to write mock functions that just return data directly ( ie. `() => ({ createPost: { id: 1 } })` instead of `() => ({ data: { createPost: { id: 1 } }})`, we could use a `LocalGraphQLError` class as a way to then check in the `LocalGraphQLClient` if the object returned is an instance of that class (then assign it to the `error` object) or just data (then assign it to `data`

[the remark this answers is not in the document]
gdorsi: Sorry, I've missed some bits in my explaination, but I'm happy that you filled the gaps and reached the same conclusions :)

gdorsi:
> that inheriting from a class just for the sake of the interface is wrong

That's true, we will try to explore other ways.

My feeling is that we should plan for a major, so we can work on reducing the `GraphQLClient` API surface.

There are other improvements that we could include in a major, e.g. revisiting the CacheAPI or rework the package structure (so we can finally kill lerna XD)

liana-pigeot:
> Let's please spend some time about thinking whether we have other options.

I've been thinking about it but can't really think of any good one, other than maybe having a `GraphQLConnection` system like I mentioned earlier which would make the local client only need to deal with responding to requests and not have to duplicate the rest of the client functionality.

Other than that I've just updated my PR with changes based on feedback, some unit tests for the `LocalGraphQLConnection`, and a way to mock errors as suggested

packages/graphql-hooks/index.d.ts
liana-pigeot: Just noticed there's already one created recently #796

liana-pigeot:
Just finished updating the typescript definitions and the documentation, the PR is ready to go once approved + CI passed

liana-pigeot:
It also seems that me running prettier with `npm run prettier` has updated the formatting of a lot of other code which I suppose wasn't formatted correctly before

simoneb (approved):
LGTM, @gdorsi can you try this out and confirm it works for you?

packages/graphql-hooks/test-d/client.test-d.ts
```diff
@@ -79,13 +85,15 @@ const result = {
...
 }
 
-expectType<<ResponseData, TGraphQLError = object>({
-  result,
-  operation
-}: {
-  result: Result<ResponseData, TGraphQLError>
-  operation: Operation<object>
-}) => void>(client.logErrorResult)
+expectType<
```
simoneb, line 88: @liana-pigeot turns out we actually have tests for types!

packages/graphql-hooks/test/unit/LocalGraphQLClient.test.js
```diff
@@ -0,0 +1,55 @@
...
+    const result = await client.request(QUERY_VARIABLES)
+    expect(result.data).toBeDefined()
+    expect(result.data.addedNumber).toBe(5)
+  })
+  it('handles error mocking', async () => {
+    const client = new LocalGraphQLClient({ localQueries })
+    const result = await client.request(QUERY_ERRORS)
+    expect(result.error).toBeDefined()
+    expect(result.error.httpError.status).toBe(404)
+  })
```
gdorsi, line 54: I think would be good to have some basic tests to check the integration with the rest of the library (e.g. with useQuery/useMutation).

Just to have something to defend us from future regressions on that side.

packages/graphql-hooks/index.d.ts
```diff
@@ -50,6 +50,19 @@ export class GraphQLClient {
   ): Promise<Result<ResponseData, TGraphQLError>>
 }
 
+export class LocalGraphQLClient extends GraphQLClient {
+  constructor(options: LocalClientOptions)
+}
+interface LocalClientOptions extends ClientOptions {
+  localQueries: LocalQueries
```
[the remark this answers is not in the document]
liana-pigeot: This one is awkward, because the parent class has `url` required the types would conflict. One problem of using inheritance here. I haven't really found any easy solution other than keeping it as it is (and it means people would have to pass a placeholder `url` value if using typescript).

This could be solved if we changed the implementation to having `GraphQLConnection` objects instead of inheriting the client, or found some other solution, but I figured the drawback of having to specify this one property isn't too bad until we can improve it

packages/graphql-hooks/index.d.ts
```diff
@@ -146,6 +159,10 @@ interface ClientOptions {
 
 type Headers = { [k: string]: string }
 
+type LocalQueries = {
+  [q: string]: (variables: object, operationName?: string) => object
```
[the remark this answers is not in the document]
liana-pigeot: I think making the interface partial still wouldn't work because it would still detect that the config passed to the constructor of `LocalGraphQLClient` isn't compatible with the constructor of `GraphQLClient`

One hacky solution could be to pass an object adding a dummy `url` prop to `super` in the constructor maybe like `super({ ...config, url: '' })`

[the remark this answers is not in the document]
gdorsi: > I think making the interface partial still wouldn't work because it would still detect that the config passed to the constructor of LocalGraphQLClient isn't compatible with the constructor of GraphQLClient

I tried locally in a typescript project and worked just fine.

Whenever we will do the migration of this codebase to Typescript, a simple casting would do the work:
`super(config as ClientOptions)`

packages/graphql-hooks/test/unit/LocalGraphQLClient.test.js
liana-pigeot: Just added one using `useQuery`

gdorsi (approved):
LGTM, well done 🚀

simoneb (approved):

PR #799, opened 2022-03-23
feat: TypeScript conversion of the codebase

### LINKED ISSUE #796: Migrate the codebase to typescript

liana-pigeot:
### What does this PR do?

(Work in progress)

Enables the progressive use of TypeScript in the repo (or at least the `graphql-hooks` package as a first step for now) while still having `js` files alongside it. This will make it easier to make changes without having to manually keep the types up to date with the code.

This PR is meant to be a proof of concept of getting the repo working with TypeScript for the `GraphQLClient`. Other files and packages can then be converted in the future

I still need to look at how to combine the type definitions from our existing `index.d.ts` file and what will be generated by actual typescript files, but TypeScript code builds so at least the `ts` code is functional.

### Related issues

Fixes #796 

### Checklist


package.json
```diff
@@ -54,5 +55,8 @@
   },
   "engines": {
     "node": "^16"
+  },
+  "dependencies": {
+    "@rollup/plugin-typescript": "^8.3.1"
```
simoneb, line 60: move to devDependencies

packages/graphql-hooks/package.json
```diff
@@ -9,6 +9,7 @@
   "scripts": {
     "clean": "rm -rf ./dist ./es ./lib",
     "build": "rollup -c",
+    "check-types": "tsc",
     "prepublishOnly": "npm run build && cp ../../README.md . && cp ../../LICENSE .",
     "tsd": "tsd"
```
simoneb, line 14: this change may make our tsd tests obsolete or largely redundant

packages/graphql-hooks/src/GraphQLClient.ts
```diff
@@ -45,12 +135,14 @@ class GraphQLClient {
...
-    this.headers = config.headers || {}
+    this.headers =
+      config.headers || (typeof Headers !== 'undefined' ? new Headers() : {})
     this.ssrMode = config.ssrMode
     this.ssrPromises = []
     this.url = config.url
     this.fetch =
-      config.fetch || (typeof fetch !== 'undefined' && fetch && fetch.bind())
+      config.fetch ||
+      (typeof fetch !== 'undefined' && fetch && fetch.bind(this))
```
simoneb, line 145: no functional change in the way we call .bind?

packages/graphql-hooks/src/GraphQLClient.ts
```diff
@@ -1,18 +1,108 @@
 import EventEmitter from 'events'
 import { extractFiles } from 'extract-files'
+import { Client } from 'graphql-ws'
```
simoneb, line 3: we should probably rename this to something more telling

simoneb (commented):
so far so good, remarkable. I left a few inline comments

packages/graphql-hooks/src/GraphQLClient.ts
liana-pigeot: I checked and the default behaviour is to use `this` from the current scope so it should be the same. TypeScript just considers not passing a parameter as invalid

simoneb: thought as much, thanks for clarifying

packages/graphql-hooks/src/GraphQLClient.ts
liana-pigeot: It was confusing for me to know what they are, especially since one is called `subscriptionCient` and this one `client` but the `subscriptionClient` one is the one that doesn't have a `subscribe` function (https://github.com/nearform/graphql-hooks/pull/799/files#diff-fa55352c78e085b512c50866f2aca7f24f34206b00aeee49f6e67250049403fbL345)

liana-pigeot:
I've updated the PR with a few things:

* tsconfig changes and other details so the type checking command now works, and I've fixed all the current type errors

* Rollup bundling changed to using `esbuild` because the rollup ts plugin keeps causing issues and is slow (it can't skip typecheck). It's a bit weird to use `esbuild` and babel at the same time, but I don't want to break existing config so I've left babel untouched. It's faster this way than using typescript anyway

* Added `ts-jest` so testing of ts files works (also I think we can write tests in typescript now) 

* Added typescript checks to CI  


Github CI seems to be broken right now, checks aren't running properly. I thought it was because I've updated the github ci file, but it seems to have happened to past commits from today before I even touched the Github ci folder... I think CI should pass, but it doesn't want to start.

liana-pigeot:
CI passes now

config/rollup.config.js
```diff
@@ -5,117 +5,125 @@ import replace from 'rollup-plugin-replace'
 import commonjs from 'rollup-plugin-commonjs'
 import { terser } from 'rollup-plugin-terser'
 import { sizeSnapshot } from 'rollup-plugin-size-snapshot'
+import esbuild from 'rollup-plugin-esbuild' // Used for TS transpiling
```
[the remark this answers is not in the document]
wilkmaia: @gdorsi I can't really say **why** it was added in the first place but there are many sources out there showing that `esbuild` is, indeed, quite faster than `babel` for this job.

A few sources are:
- https://github.com/evanw/esbuild
- https://github.com/privatenumber/esbuild-loader/discussions/138
- https://datastation.multiprocess.io/blog/2021-11-13-benchmarking-esbuild-swc-typescript-babel.html

Far from being a benchmark in itself, I've run some quick tests locally and building `graphql-hooks` with `babel` takes ~6 seconds on my machine. With `esbuild` it takes ~4 seconds.

-----

That aside, any particular issue why you think `babel` should be used instead?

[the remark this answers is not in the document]
gdorsi: We are already using babel for the es5 compilation, so to support the typescript compilation would be enough to add the typescript plugin.

More than solving a real problem, I was curious if adding the typescript plugin on babel costs us more than adding esbuild.

Thanks for checking!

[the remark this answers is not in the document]
liana-pigeot: The main reason I added esbuild is that babel isn't used everywhere in this. one of the build types doesn't use babel and I assumed this was intentional to provide non transpiled files.

So instead of adding babel on things that didn't have it, I just make everything go through esbuild to strip the typescript.

I also remember having another specific issue when I tried with babel, so I ended up switching to esbuild which solves the problem easily without issues and compiles fast

There's no documentation on why the build was was like that in the first place so it's hard to tell what's the best thing to do, but the current setup works

rp4rk (approved):
This LGTM, there is a small linting warning but otherwise I think it suitably meets the goals stated in the PR description. The tests around types might be useful for a while longer to ensure no unexpected divergence while iterated over imo. Nice work 👍

gdorsi (approved):
LGTM, good job 🚀

PR #828, opened 2022-04-25
test: convert tests to TypeScript

ianlovell-nearform:
- Convert tests to TypeScript


jest.config.js
```diff
@@ -23,16 +22,14 @@ const projects = [
...
-      testMatch: ['<rootDir>/packages/graphql-hooks-memcache/**/*.test.[jt]s']
+      displayName: 'graphql-hooks-memcache'
     }
   },
   {
     ...commonConfig,
     ...{
       roots: ['./packages/graphql-hooks-ssr'],
-      displayName: 'graphql-hooks-ssr',
-      testMatch: ['<rootDir>/packages/graphql-hooks-ssr/**/*.test.[jt]s']
```
simoneb, line 35: why are these not needed anymore?

ianlovell-nearform: the default `testMatch` covers `spec.*`/`test.*` and `js`/`ts`/`jsx`/`tsx` extensions as well as the `__tests__` folder - as `roots` is set, only the appropriate folder will be searched

simoneb: and it only navigate in `roots` then? that was my main question really, to avoid each of these to run ALL the tests in the repo

ianlovell-nearform: yes that's correct 👍 18 test suite passes/178 test passes in both `master` and this PR

simoneb (approved):

rp4rk (approved):
This looks good to me :)

simoneb:
let's merge!

Issue #838, opened 2022-04-28
Typescript definitions for useQuery, useSubscription, useClientRequest aren't exported properly


ekynoxe:
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: 5.11.1

### Description

The exported types for `useQuery`, `useSubscription`, `useClientRequest` all use `any` as parameter types, and the corresponding "declared function" in `packages/graphql-hooks/src/types/common-types.ts` aren't used.

For example, `useQuery` is exported as

```
declare function useQuery(query: any, opts?: {}): any;
export default useQuery;
```

instead of

```
import { UseQueryOptions, UseQueryResult } from './types/common-types';
declare function useQuery<ResponseData = any, Variables = object, TGraphQLError = object>(query: string, opts?: UseQueryOptions<ResponseData, Variables>): UseQueryResult<ResponseData, Variables, TGraphQLError>;
export default useQuery;
```

### How to reproduce

Simply build the packages, or try to use `useQuery` in a project (We're using it in a client project and found the type issue whilst trying to upgrade to the latest version).

### Suggested solution (optional)

Could migrating all files to TS help in this?

PR #839, opened 2022-04-28
fix: update apq middleware build

### LINKED ISSUE #819: Is APQ working?
### Package

`graphql-hooks`

### Description

We recently release support for APQ and during the publishing of the package several warnings were generated:

```
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
@aws-crypto/sha256-universal (imported by src\middlewares\apqMiddleware.js)
(!) Mixing named and default exports
https://rollupjs.org/guide/en/#outputexports
The following entry modules are using named and default exports together:
src\middlewares\apqMiddleware.js

Consumers of your bundle will have to use chunk['default'] to access their default export, which may not be what you want. Use `output.exports: 'named'` to disable this warning
created lib/middlewares/apqMiddleware.js in 260ms
```

We should try whether using this middleware from an app actually works.

ianlovell-nearform:
- Replaces `@aws-crypto/sha256-universal` with `@aws-crypto/sha256-browser`
- `APQMiddleware` now a named export
- Replaces TestCafe with Cypress
- Runs acceptance tests on CI

resolves #819


simoneb:
@rp4rk is looking into https://github.com/nearform/graphql-hooks/issues/838 which has also to do with typings, let's make sure we come to an organic solution

rp4rk:
This LGTM so far

examples/create-react-app/src/App.js
```diff
@@ -1,5 +1,7 @@
 import React, { useState } from 'react'
 import { GraphQLClient, ClientContext } from 'graphql-hooks'
+// eslint-disable-next-line no-unused-vars
```
rp4rk, line 3: Was including this intended?

packages/graphql-hooks/rollup.config.js
```diff
@@ -14,5 +15,12 @@ export default generateRollupConfig({
     format: 'cjs',
     indent: false
   },
-  plugins: [esbuild(), babel(), sizeSnapshot()]
+  plugins: [
+    esbuild(),
+    replace({
```
rp4rk, line 20: I can't find reference to `process.env.NODE_ENV` outside of some configuration files, is this vestigial?

rp4rk (approved):
Couple of things I've noticed but I don't think any of them are huge blockers at all if they prove to be

examples/create-react-app/src/App.js
ianlovell-nearform: yeah, the APQ issue with `crypto` not being polyfilled with webpack 5 only surfaces when the middleware is imported, once #847 is complete, that example could be used for the acceptance tests

PR #840, opened 2022-04-28
Chore/export definitions correctly

rp4rk:
### What does this PR do?

Converts remaining JS code to TS in the graphql-hooks project, in order to correctly export type definitions. Previously, declarations being made on the global scope weren't being considered due to declarations being generated directly from the JS files.

Note: tests all pass, but there are some functional changes in this PR, notably certain functions now throw errors if not provided with an appropriate client to use for requests.

### Related issues

Closes https://github.com/nearform/graphql-hooks/issues/838

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/types/common-types.ts
+++ b/packages/graphql-hooks/src/types/common-types.ts
@@ -92,28 +92,6 @@ declare function useClientRequest<
   ResetFunction
 ]
 
-declare function useQuery<
-  ResponseData = any,
-  Variables = object,
-  TGraphQLError = object
->(
-  query: string,
-  options?: UseQueryOptions<ResponseData, Variables>
-): UseQueryResult<ResponseData, Variables, TGraphQLError>
-
-declare function useManualQuery<
-  ResponseData = any,
-  Variables = object,
-  TGraphQLError = object
->(
-  query: string,
-  options?: UseClientRequestOptions<ResponseData, Variables>
-): [
-  FetchData<ResponseData, Variables, TGraphQLError>,
-  UseClientRequestResult<ResponseData, TGraphQLError>,
-  ResetFunction
-]
-
 declare function useMutation<
   ResponseData = any,
   Variables = object,
--- a/packages/graphql-hooks/src/types/common-types.ts
+++ b/packages/graphql-hooks/src/types/common-types.ts
@@ -79,32 +79,6 @@ declare class LocalGraphQLError<TGraphQLError = object>
   constructor(error: APIError<TGraphQLError>)
 }
 
-declare function useClientRequest<
-  ResponseData = any,
-  Variables = object,
-  TGraphQLError = object
->(
-  query: string,
-  options?: UseClientRequestOptions<ResponseData, Variables>
-): [
-  FetchData<ResponseData, Variables, TGraphQLError>,
-  UseClientRequestResult<ResponseData, TGraphQLError>,
-  ResetFunction
-]
-
-declare function useMutation<
-  ResponseData = any,
-  Variables = object,
-  TGraphQLError = object
->(
-  query: string,
-  options?: UseClientRequestOptions<ResponseData, Variables>
-): [
-  FetchData<ResponseData, Variables, TGraphQLError>,
-  UseClientRequestResult<ResponseData, TGraphQLError>,
-  ResetFunction
-]
-
 export interface SubscriptionRequest {
   query: string
   variables: object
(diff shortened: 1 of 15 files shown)


packages/graphql-hooks/src/types/common-types.ts
```diff
@@ -79,54 +79,6 @@ declare class LocalGraphQLError<TGraphQLError = object>
   constructor(error: APIError<TGraphQLError>)
 }
 
-declare function useClientRequest<
```
rp4rk, line 82: These declarations should now be inline with the appropriate functions

packages/graphql-hooks/src/useClientRequest.ts
```diff
@@ -56,13 +63,13 @@ function reducer(state, action) {
...
 function useDeepCompareCallback(callback, deps) {
-  const ref = React.useRef()
+  const ref = React.useRef<DependencyList>()
 
   if (!dequal(deps, ref.current)) {
     ref.current = deps
   }
 
-  return React.useCallback(callback, ref.current)
+  return React.useCallback(callback, ref.current || [])
```
simoneb, line 72: this is a suspicious change

packages/graphql-hooks/src/useClientRequest.ts
```diff
@@ -82,18 +100,25 @@ function useClientRequest(query, initialOpts = {}) {
 
   const contextClient = React.useContext(ClientContext)
   const client = initialOpts.client || contextClient
+
+  if (client === null) {
```
simoneb, line 104: what if it's undefined?

packages/graphql-hooks/src/useClientRequest.ts
```diff
@@ -188,8 +215,8 @@ function useClientRequest(query, initialOpts = {}) {
 
         const actionResult = { ...result }
         if (revisedOpts.useCache) {
-          actionResult.useCache = true
-          actionResult.cacheKey = revisedCacheKey
+          ;(actionResult as any).useCache = true
+          ;(actionResult as any).cacheKey = revisedCacheKey
```
simoneb, line 219: ugly syntax, can we do any better?

packages/graphql-hooks/src/useClientRequest.ts
```diff
@@ -82,18 +100,25 @@ function useClientRequest(query, initialOpts = {}) {
...
     variables: initialOpts.variables,
     operationName: initialOpts.operationName,
     persisted: initialOpts.persisted
-  }
+  } as any
 
   if (
     initialOpts.persisted ||
-    (client.useGETForQueries && !initialOpts.isMutation)
+    (client?.useGETForQueries && !initialOpts.isMutation)
```
simoneb, line 121: if we assert the client is there, why do we need a conditional chaining here? (and below)

packages/graphql-hooks/test/unit/useManualQuery.test.ts
```diff
@@ -10,10 +10,10 @@ const TEST_QUERY = `query Test($limit: Int) {
 
 describe('useManualQuery', () => {
   it('calls useClientRequest with useCache set to true & options', () => {
-    useManualQuery(TEST_QUERY, { option: 'option' })
+    useManualQuery(TEST_QUERY, { isMutation: true })
     expect(useClientRequest).toHaveBeenCalledWith(TEST_QUERY, {
       useCache: true,
-      option: 'option',
+      isMutation: true,
```
simoneb, line 16: this was changed because since it's typed we can't just provide any options?

rp4rk: Yeah the type is restrictive here

packages/graphql-hooks/src/useClientRequest.ts
rp4rk: I believe it can't be, if `initialOpts.client` is undefined here it will instead fall to `contextClient` which can either be a client or `null`.

packages/graphql-hooks/src/useClientRequest.ts
rp4rk: Pushed an alternative approach to it that doesn't fallback to creating an empty list.

simoneb: not saying it is, but we have a history of introducing seemingly innocuous changes which turn then out to be breaking. so let's try to limit then as much as possible unless we're 100% sure

ianlovell-nearform (approved):
LGTM

rp4rk:
Screenshot 2022-04-29 at 11 06 45

For a bit of confidence, here is the `useQuery` type hinting visible in the IDE, compared to the current `master` branch,

Screenshot 2022-04-29 at 11 07 36

simoneb (approved):

PR #848, opened 2022-05-03
feat: add apq example

### LINKED ISSUE #847: Add APQ example
- Add an example showing the APQ middleware running against an API

ianlovell-nearform:
- Adds persisted queries example
- Tests APQ middleware through Cypress
- Fixes `Buffer` polyfill issue with webpack 5

resolves #847


rp4rk (approved):

lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "packages": ["packages/*", "examples/*"],
+  "packages": ["packages/*", "examples/*", "examples/persisted-queries/*"],
```
simoneb, line 2: why  this change? doesn't examples/* capture this already?

examples/subscription/package.json
```diff
@@ -20,7 +20,7 @@
     "fastify-cors": "^6.0.2",
     "fastify-gql": "^5.0.0",
     "fastify-static": "^4.5.0",
-    "graphql": "^16.0.1",
+    "graphql": "^15.4.0",
```
simoneb, line 23: why this downgrade?

cypress.json
```diff
@@ -1,5 +1,4 @@
 {
   "defaultCommandTimeout": 6000,
-  "video": false,
-  "baseUrl": "http://localhost:3000"
```
simoneb, line 4: why do we need them to run on different ports? I don't expect we need to run them concurrently, do we?

packages/graphql-hooks/src/middlewares/apqMiddleware.ts
```diff
@@ -1,4 +1,5 @@
 import { Sha256 } from '@aws-crypto/sha256-browser'
+import { Buffer } from 'buffer'
```
simoneb, line 2: do we need this? it's a fairly large package

simoneb (commented):
See inline comments. Plus, where is this example running?

packages/graphql-hooks/src/middlewares/apqMiddleware.ts
ianlovell-nearform: this is not pollyfilled with webpack 5, so the APQ middleware throws an exception in a browser environment, one option would be to require a `sha256` function as an option to the middleware and let the consumer choose, another option would be to separate the apq middleware to a separate package, which may also provide better encapsulation (currently imported from `graphql-hooks/lib/middlewares/apqMiddleware`)

cypress.json
ianlovell-nearform: there was an issue with cypress not shutting down the first client (create-react-app) before it attempting to run the second client (persisted-queries) within `acceptance-tests.yml`

examples/subscription/package.json
ianlovell-nearform: this was to fix an issue with npm's peer dependency check when using `apollo-server` and the existing `json-graphql-server` both requiring `graphql`

lerna.json
ianlovell-nearform: the `persisted-queries` folder has a `client` and `server` package subfolders, so they are not covered with the existing glob, was hitting https://github.com/lerna/lerna/issues/2127 with other setups

simoneb (approved):

PR #852, opened 2022-05-05
fix: update jest and formdata test

### LINKED ISSUE #832: Undo workaround for formdata-node export issue
### Description

In https://github.com/nearform/graphql-hooks/pull/760/files#diff-ac30df9c379259c8407e252c5de7ae4de97119f9d903450d1c5454c3810cead8R8 we introduced a workaround for https://github.com/octet-stream/form-data/issues/50.

Now that the issue seems to be solved by upgrading jest, we should try to remove the workaround by upgrading jest (which is broken btw https://github.com/nearform/graphql-hooks/pull/830).

ianlovell-nearform:
- Updates jest, babel-jest, and ts-jest
- Updates formdata file test to use jest node environment

resolves #832


raul-saez (approved):

packages/graphql-hooks/test/unit/GraphQLClient.test.ts
```diff
@@ -9,9 +9,6 @@ import {
   createMockSubscriptionClient
 } from '../utils'
 
-// workaround for https://github.com/octet-stream/form-data/issues/50
-const { fileFromPathSync } = require('formdata-node/lib/cjs/fileFromPath')
```
simoneb, line 13: I think we have an example of using this in the README. Are we sure we don't want to keep using the real file for consistency with the README or even just to make sure it actually works with formdata-node? Ideally another change should be included in this PR, either or both the README or the code, to cover both scenarios of files from the file system and blobs/files in memory

simoneb (changes requested):

packages/graphql-hooks/test/unit/GraphQLClient.test.ts
ianlovell-nearform: Have reinstated the file from disk example, the jest environment needed to be updated to node for that case, so the tests are currently split into jsdom/node folders

Issue #920, opened 2022-09-16
All the test suites are failing


radomird:
### Package

`graphql-hooks`
`graphql-hooks-ssr`
`graphql-hooks-memcache`

### Description

When running `npm run test`, from the root of the project, all the unit test fail with a similar message (these are just a few examples):
```
error TS2339: Property 'toBe' does not exist on type 'Assertion'.
error TS2339: Property 'toHaveBeenCalled' does not exist on type 'Assertion'.
error TS2339: Property 'rejects' does not exist on type 'Assertion'.
```
### How to reproduce

Run `npm run test`.

### Suggested solution (optional)

Most likely the reason for the failures is a misconfiguration/or missing some config. 

Part of fixing this issue should be also to remove the `tsd` dependency and type tests (located in `/packages/graphql-hooks/test-d/` folder) as they are **most likely not needed anymore** since we are using `tsc` to check the types (check this [comment](https://github.com/nearform/graphql-hooks/pull/799#discussion_r833635348) and [PR](https://github.com/nearform/graphql-hooks/pull/799) for more information).

simoneb:
it doesn't seem to be the case to me. if you go to the repo root you'll see that the last commit on master has a green checkmark

radomird:
> it doesn't seem to be the case to me. if you go to the repo root you'll see that the last commit on master has a green checkmark

Yeah I saw that, but I can consistently reproduce it locally on `master` branch. I'm looking into it.

simoneb:
if it succeeds in CI the most sensible assumption is that your local environment has problems, e.g. node version or other such things which deviate from how CI is configured

radomird:
> if it succeeds in CI the most sensible assumption is that your local environment has problems, e.g. node version or other such things which deviate from how CI is configured

Well the thing is it fails in the CI as well (see the latest [PR build](https://github.com/nearform/graphql-hooks/actions/runs/3080743477/jobs/4978455546)). The issue is that Cypress global TypeScript type conflicts with Jest `expect` (see https://github.com/cypress-io/cypress/issues/22059).

simoneb:
I still don't quite understand why the build is green on master though

PR #921, opened 2022-09-19
chore: removed tsd and type tests

### LINKED ISSUE #922: [chore] Remove `tsd` library and type tests
### Description

Removes the tsd library and all the type tests from the graphql-hooks package. Reason for removing is that hte whole library was re-written to use Typescript and the check-types script will check all the types, so running tsd became obsoloete (see the following https://github.com/nearform/graphql-hooks/pull/799 and https://github.com/nearform/graphql-hooks/pull/799#discussion_r833635348).

Related PR:
https://github.com/nearform/graphql-hooks/pull/919

radomird:
### What does this PR do?

Removes the `tsd` library and all the type tests from the `graphql-hooks` package. Reason for removing is that hte whole library was re-written to use Typescript and the `check-types` script will check all the types, so running `tsd` became obsoloete (see the following [PR](https://github.com/nearform/graphql-hooks/pull/799) and [comment](https://github.com/nearform/graphql-hooks/pull/799#discussion_r833635348)).

### Related issues

https://github.com/nearform/graphql-hooks/pull/919

Closes #922


package.json
```diff
@@ -9,7 +9,6 @@
     "build:packages": "lerna run build --no-private",
     "build:site": "lerna run build --scope=create-react-app",
     "postinstall": "lerna bootstrap --hoist --no-ci && npm run build:packages",
-    "tsd": "lerna run tsd",
```
simoneb, line 12: this command assumes that there's a tsd script in at least some of the monorepo package.json files. we should remove it from there too I guess

radomird: Yes, I've checked all of them and there weren't in any of the packages...

simoneb: then this command was already doing nothing :D https://github.com/nearform/graphql-hooks/actions/runs/2950012855/jobs/4714237058#step:8:12

radomird:
The build fails because of this issue: https://github.com/nearform/graphql-hooks/issues/920.
After resolving that we can proceed with merging this PR.

ilteoood (approved):

PR #923, opened 2022-09-19
fix: updated config files to ignore cypress types on global level

### LINKED ISSUE #920: All the test suites are failing

radomird:
Closes #920


cypress/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
+  "extends": "../tsconfig",
   "compilerOptions": {
-    "target": "es5",
-    "lib": ["es5", "dom"],
     "types": ["cypress", "@testing-library/cypress"]
   },
-  "include": ["**/*.ts"]
+  "include": ["**/*.ts"],
+  "exclude": [],
```
ilteoood, line 7: Is this exclude necessary?

ilteoood (approved):

tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
   "compilerOptions": {
-    "lib": ["es2021", "dom"],
-    "module": "esnext",
-    "target": "es2021",
+    "baseUrl": ".",
+    "outDir": "lib",
```
simoneb, line 4: this is suspicious. why are we changing the output directory of the transpilation? unless we know what we're doing here, this is potentially affecting the output of the transpilation and what is published in the npm packages. it's unusual for the output of transpilation to be in a lib folder

cypress/tsconfig.json
radomird: Absolutely, since we are extending the root `tsconfig.json` and that one excludes `cypress` we need to overwrite the exclude prop for the cypress to run.

tsconfig.json
radomird: The reason why I've added the `outDir` option is because I kept getting into the following error, on every file that was transpiled: 
```
error TS5055: Cannot write file '...' because it would overwrite input file
```
Which makes sense, as TypeScript was trying to write JS files into the same directory they were discovered from. Introducing a dir where it will write these files will solve this issue. The reason why it has `lib` as a value is that I just copied it from the `/packages/graphql-hooks/tsconfig.json`.

However - you have a good point for affecting the output dir (although running the `build` script will generate the outputs where they should be originally, no changes there - this would just create a new folder if you would run `tsc` command in the root folder) and after some investigation - we can resolve the above mentioned issue also by adding ```noEmit: true``` to the compiler options (instead of `outDir`).

simoneb:
@radomird @ilteoood can you please decide together which one of the PRs to address #920 should go in? I believe that if they achieve the same, #924 is a simpler option

radomird:
@ilteoood Let's use your PR https://github.com/nearform/graphql-hooks/pull/924 
Closing this PR.

Issue #930, opened 2022-10-05
setQueryData cannot receive variables or operationName.


Twipped:
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: 6.0.0

### Description

The new `setQueryData` function released in 6.0.0 only accepts a query string for targeting query data. That means you cannot update query results if an operationName or any variables are provided for the query.

See line https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks/src/GraphQLClient.ts#L429

### Suggested solution (optional)

Change `setQueryData` to accept an operation object as the first argument.

lasmil:
I am checking this.

simoneb:
@Twipped can you check if the linked pull request addresses your request?

Twipped:
Spot on, thank you!

PR #931, opened 2022-10-05
fix: 930: use operation object in setQueryData

lasmil:
This PR will enable the setQueryData and invalidateQuery functions to receive operation object when being called.
Backward compatibility is met because we have changed the definition of these functions.

Closes: #930

### DIFF
--- a/packages/graphql-hooks/src/GraphQLClient.ts
+++ b/packages/graphql-hooks/src/GraphQLClient.ts
@@ -413,11 +413,13 @@ class GraphQLClient {
     }
   }
 
-  invalidateQuery(query: string): void {
-    const cacheKey = this.getCacheKey({ query })
+  invalidateQuery(query: Operation | string): void {
+    const cacheKeyProp = (typeof query === 'string' ? { query } : query) as Operation
+
+    const cacheKey = this.getCacheKey(cacheKeyProp)
     if (this.cache && cacheKey) {
       this.removeCache(cacheKey)
-      this.request({ query })
+      this.request(cacheKeyProp)
         .then(result => {
           this.mutationsEmitter.emit(Events.DATA_INVALIDATED, result)
         })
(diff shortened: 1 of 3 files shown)


packages/graphql-hooks/test-node/GraphQLClient.test.ts
```diff
@@ -162,8 +162,113 @@ describe('with files in Node JS', () => {
...
+    expect(cacheMock.get).toBeCalledWith({ query: validQuery })
+
+    expect(updaterMock).toBeCalledTimes(2)
+    expect(updaterMock).toBeCalledWith(resultMock)
+
+    expect(mutationsEmitterMock.emit).toBeCalledTimes(2)
+    expect(mutationsEmitterMock.emit).toBeCalledWith(Events.DATA_UPDATED, {
+      data: resultMock
+    })
+  })
```
mahenrique94, line 273: Maybe you should add one more test using `variables` property

lasmil: Done

mahenrique94 (approved):
👏 👏👏

packages/graphql-hooks/src/GraphQLClient.ts
```diff
@@ -425,8 +425,11 @@ class GraphQLClient {
     }
   }
 
-  setQueryData(query: string, updater: (oldState?: any) => any): void {
-    const cacheKey = this.getCacheKey({ query })
+  setQueryData(query: string, updater: (oldState?: any) => any, operation: Operation | {} = {}): void {
```
simoneb, line 428: let's make the query argument dynamic so that if it's a string we assume it's the query, if it's an object we treat it as an operation. let's also update the invalidatQuery method to do the same.

simoneb (approved):

PR #953, opened 2022-11-02
fix: useManualQuery queryFn drilled to child component and invoked on mount does not toggle loading boolean

### LINKED ISSUE #949: useManualQuery queryFn drilled to child component and invoked on mount does not toggle loading boolean
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: 5.11.4
- `Node` version: 14.20.0
- `react` version: 18.2.0
- Browser: Chrome/Firefox

### Description

If using `useManualQuery`, I've found that when the queryFn is drilled to a child component as a prop, the loading state on the query is not toggled properly, even though the data is retrieved. I believe this might be due to the following snippet running on mount:

```jsx
  // NOTE: state from useReducer is only initialState on the first render
  // in subsequent renders the operation could have changed
  // if so the state would be invalid, this effect ensures we reset it back
  const stringifiedCacheKey = JSON.stringify(cacheKey)
  React.useEffect(() => {
    if (!initialOpts.updateData) {
      // if using updateData we can assume that the consumer cares about the previous data
      dispatch({ type: actionTypes.RESET_STATE, initialState })
    }
  }, [stringifiedCacheKey]) // eslint-disable-line react-hooks/exhaustive-deps
  ```
  
 In the case of a manual query, the initial value of loading is `false`, which could be preventing the boolean from being toggled to `true`, although I'm not certain why this behavior differs depending on whether the queryFn is invoked in a child component or the component which contains the useManualQuery invocation. Potentially the state updates are being batched in a different manner or order in this case.

### How to reproduce

Attempt to invoke useManualQuery's queryFn after passing as a prop, on mount, as seen in the following code snippet:

https://codesandbox.io/s/musing-parm-vo2uz9?file=/src/Fetcher.tsx

We can see that the loading value never toggles from false to true, but the data is successfully retrieved. This seems to be the case both with and without React StrictMode.

Note that commenting the useEffect in `Child` out and uncommenting the useEffect in `Fetcher` causes the query to behave as expected regarding the loading boolean.

The state reset is skipped when an `updateData` option is passed. So passing
```ts
updateData: (_data: object, nextData: object) => nextData,
```
Seems to allow the loading boolean to toggle as expected. This doesn't seem to be intended, and is more so taking advantage of an implementation detail, and so I'm wondering if the reason for this issue can be elaborated on, and a better solution proposed.

Thanks, have a great day.

andreaforni:
### What does this PR do?
This PR fixes #949 bug. To do that, it adds a check in the `useClientRequest` reducer not to reset the state if we are loading data.

It also adds a new integration test file for `useManualQuery`, it contains the same tests defined in `useQuery` plus new ones to keep track of this bug (the ones inside the "_useManualQuery with nested components_" describe block).

### Related issues 
#949 

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/useClientRequest.ts
+++ b/packages/graphql-hooks/src/useClientRequest.ts
@@ -20,6 +20,10 @@ const actionTypes = {
 function reducer(state, action) {
   switch (action.type) {
     case actionTypes.RESET_STATE:
+      // Do not reset the state if it's loading
+      if (state.loading) {
+        return state
+      }
       return action.initialState
     case actionTypes.LOADING:
       // if the previous action resulted in an error - refetch should clear any errors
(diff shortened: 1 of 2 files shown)


packages/graphql-hooks/src/useClientRequest.ts
```diff
@@ -20,6 +20,10 @@ const actionTypes = {
 function reducer(state, action) {
   switch (action.type) {
     case actionTypes.RESET_STATE:
+      // Do not reset the state if it's loading
+      if (state.loading) {
+        return state
+      }
```
andreaforni, line 26: This is the fix, the other changes in this file are generated by prettier.

simoneb (approved):

PR #992, opened 2023-01-05
fix: removed src folder from graphql-hooks and graphql-hooks-memcache package.json files entry

### LINKED ISSUE #991: Src folder with typescript files shipped in npm package
### Package
`graphql-hooks`

### Environment
- `graphql-hooks` version: 6.0.5
- `react` version: 18
- `webpack`version: 5

### Description
My react 18 project is using a plugin which in turn makes use of the grapqhql-hooks library. Since the recent update to webpack 5 an error is thrown during the build process:

```
ERROR in ../node_modules/graphql-hooks/src/events.ts 1:7
  Module parse failed: Unexpected token (1:7)
  You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
  > export enum Events {
  |   DATA_INVALIDATED = 'DATA_INVALIDATED',
  |   DATA_UPDATED = 'DATA_UPDATED'`
```
For some reason the src folder is shipped with the graphql-hooks library and webpack tries to load the typescript files contained in it. I don't see a reason why the src folder should be shipped with the package.

### How to reproduce
I couldn't reproduce the error in a fresh react app. However, the src folder is contained in node_modules/graphql-hooks if you just create a new project and install the graphql-hooks library from npm:
`npm i graphql-hooks@6.0.5` 

### Suggested solution
It seems like there is a reference to one or more of the typescript files from the type definitions, I couldn't figure out though, which one. Furthermore the folder structure in the /lib/types folder seems weird. Besides the _d.ts files it includes another types folder (/lib/types/types) that only contains the common-types.d.ts file. Could you please have a look if there is an unintended reference?

stearm:
### What does this PR do?

This PR does not ship any `src` directories, to prevent TS files parsing errors. It resolves #991 since we don't have any reproducible example.

### Related issues

#991 

### Checklist

### DIFF
--- a/packages/graphql-hooks-memcache/package.json
+++ b/packages/graphql-hooks-memcache/package.json
@@ -16,7 +16,6 @@
     "dist",
     "es",
     "lib",
-    "src",
     "index.d.ts"
   ],
   "keywords": [
--- a/packages/graphql-hooks/package.json
+++ b/packages/graphql-hooks/package.json
@@ -18,7 +18,6 @@
     "dist",
     "es",
     "lib",
-    "src",
     "index.d.ts"
   ],
   "keywords": [


simoneb:
Considering that we still don't have a working repro for the related issue, I would suggest that we keep this open until we have one so we try to address and fix the issue. If the repro doesn't come timely, we can consider merging this anyway

simoneb (approved):
LGTM but please link the issue properly in the PR before merging

PR #993, opened 2023-01-09
feat: refetch a query only if mutation doesn't fail (refetchAfterMutation option)

### LINKED ISSUE #988: Way to refetchAfterMutations only if mutation succeeds
### Description

Right now, if you utilize refetchAfterMutations, it will refetch even if the mutation fails.

### Suggested implementation

It would be better if it only refetched on success! If the mutation fails, we can reasonable assume that the query doesn’t need to be refetched.

stearm:
### What does this PR do?
With this PR, a query is refetched only if the relative mutation (linked using `refetchAfterMutation` option) is successful.
I also renamed `conditionalRefetch` to `afterConditionsCheckRefetch` since it's not only a conditional refetch anymore, it also check for errors.

Introduced `refetchAfterMutations` as an optional parameter in [RefetchAfterMutationItem](https://github.com/nearform/graphql-hooks/pull/993/files#diff-d1d76b2c9ec609544e715dedb4b77efe053d6798016eeebc562a64dd3765cc79R187) type to determine whether or not to refetch the query if the mutation fails.

### Related issues

It resolves #988.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,7 +255,7 @@ This is a custom hook that takes care of fetching your query and storing the res
     - If it's a string, it's the mutation string
     - If it's an object then it has properties mutation and filter
       - `mutation`: String - The mutation string
-      - `refetchOnMutationError`: boolean (optional, fallback to `true`) - It indicates whether the query must be re-fetched due to a mutation error
+      - `refetchOnMutationError`: boolean (optional, fallback to `true`) - It indicates whether the query must be re-fetched if the mutation returns an error
       - `filter`: Function (optional) - It receives mutation's variables as parameter and blocks refetch if it returns false
     - If it's an array, the elements can be of either type above
 
--- a/README.md
+++ b/README.md
@@ -255,7 +255,7 @@ This is a custom hook that takes care of fetching your query and storing the res
     - If it's a string, it's the mutation string
     - If it's an object then it has properties mutation and filter
       - `mutation`: String - The mutation string
-      - `refetchOnMutationError`: boolean (optional, fallback to `true`) - It indicates whether the query must be re-fetched if the mutation returns an error
+      - `refetchOnMutationError`: boolean (optional, defaults to `true`) - It indicates whether the query must be re-fetched if the mutation returns an error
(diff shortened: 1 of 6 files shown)


simoneb (commented):
@stearm shall we introduce this behavior as an option instead? otherwise this is going to be a breaking change

stearm:
> @stearm shall we introduce this behavior as an option instead? otherwise this is going to be a breaking change

Sure, you can look at ef39fa0. Thank you!

README.md
```diff
@@ -255,6 +255,7 @@ This is a custom hook that takes care of fetching your query and storing the res
     - If it's a string, it's the mutation string
     - If it's an object then it has properties mutation and filter
       - `mutation`: String - The mutation string
+      - `refetchOnMutationError`: boolean (optional, fallback to `true`) - It indicates whether the query must be re-fetched due to a mutation error
```
simoneb, line 258:
```suggestion
      - `refetchOnMutationError`: boolean (optional, defaults to `true`) - It indicates whether the query must be re-fetched due to a mutation error
```

packages/graphql-hooks/src/createRefetchMutationsMap.js
```diff
@@ -19,9 +19,14 @@ export default function createRefetchMutationsMap(refetchAfterMutations) {
...
     } else if (paramType === 'object') {
-      const { filter, mutation } = mutationInfo
+      const { filter, mutation, refetchOnMutationError } = mutationInfo
 
-      result[mutation] = { filter }
+      result[mutation] = {
+        filter,
+        refetchOnMutationError:
+          typeof refetchOnMutationError === 'undefined' ||
+          refetchOnMutationError
```
simoneb, line 28: why this logic? can't we assume the user will provide a sensible value? if this is to apply the default value, let's just default it to true when destructuring it above

stearm: you're right, a default fallback is enough.

PR #997, opened 2023-01-17
973 / Fix APIError Type

### LINKED ISSUE #973: Fix APIError Type
### Package
`graphql-hooks`

### Environment

- `graphql-hooks` version: 6.0.3

### Description

The passed generic type (TGraphQLError) is not used inside the APIError interface [here](https://github.com/nearform/graphql-hooks/blob/4ad8fa015bd98dd8d428e63061c79c006369623e/packages/graphql-hooks/src/types/common-types.ts#L132)
```export interface APIError<TGraphQLError = object> {
    fetchError?: Error;
    httpError?: HttpError;
    graphQLErrors?: GraphQLResponseError[];
}```

kephail:
Remove the unused generic type `TGraphQLError` within the interface 

### Related issues

fixes #973 

Spent a bit of time today trying to update this generic but found myself facing issues with types further down the chain where we are specifying `any` or `object` such as in: https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks/src/GraphQLClient.ts#L157

In the end I decided it was easier for this particular issue to just remove the unused generic. But I wonder if we may want another issue opening to type these other areas more strongly rather than relying on `any` and `object`.

### Checklist

- [ ] I have checked the [contributing document](../blob/master/CONTRIBUTING.md)

### DIFF
--- a/packages/graphql-hooks/src/GraphQLClient.ts
+++ b/packages/graphql-hooks/src/GraphQLClient.ts
@@ -2,24 +2,24 @@ import EventEmitter from 'events'
 import { extractFiles } from 'extract-files'
 import { Client as GraphQLWsClient } from 'graphql-ws'
 import { SubscriptionClient } from 'subscriptions-transport-ws'
-import canUseDOM from './canUseDOM'
-import isExtractableFileEnhanced from './isExtractableFileEnhanced'
 import Middleware from './Middleware'
+import canUseDOM from './canUseDOM'
+import { Events } from './events'
+import isExtractableFileEnhanced from './isExtractableFileEnhanced'
 import type {
-  UseClientRequestOptions,
   Cache,
+  CacheKeyObject,
   ClientOptions,
   FetchFunction,
   GenerateResultOptions,
+  GraphQLResponseError,
   OnErrorFunction,
   Operation,
   RequestOptions,
   Result,
-  GraphQLResponseError,
-  CacheKeyObject
+  UseClientRequestOptions
 } from './types/common-types'
 import { pipeP } from './utils'
-import { Events } from './events'
 
 class GraphQLClient {
   url: string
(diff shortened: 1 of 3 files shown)


eugenio-oddone (approved):
Looks good to me, APIError still typed. definitely agree with this:

> In the end I decided it was easier for this particular issue to just remove the unused generic. But I wonder if we may want another issue opening to type these other areas more strongly rather than relying on any and object.

simoneb:
@kephail if we merge this change, will it be a breaking change (in terms of typings) for existing users?

cuongtq17 (approved):

toomuchdesign:
> @kephail if we merge this change, will it be a breaking change (in terms of typings) for existing users?

Since `LocalGraphQLError` is public and its generic support was removed, then this PR might be considered a breaking change. Existing codebases passing a generic to `LocalGraphQLError` would get a type error.

simoneb:
thanks @toomuchdesign . based on this @kephail do you think there's a way to introduce this change in a non-breaking way?

toomuchdesign:
The only solution I can think of is keeping an unused generic support in user facing types.

simoneb:
And just for clarity, the generics didn't make any sense at all or they would make sense, if we were using them more correctly in our types?

toomuchdesign:
All the removed generics ended up being passed down as [`APIError<TGraphQLError = object>`](https://github.com/nearform/graphql-hooks/blob/51e4a25734ecd45bd34e844f896e1abd1498037a/packages/graphql-hooks/src/types/common-types.ts#L132) generic.

Not sure what the original intentions were, maybe provide a custom type for `graphQLErrors` prop? If so we might use the generic and introduce a breaking change that resembles more as a fix.

simoneb:
Superseded by #1013

PR #1008, opened 2023-02-03
feat(example): create abortcontroller example

### LINKED ISSUE #1003: Allow abort of fetch via AbortController
### Description

It's possible there is a way to handle this case already, and I simply haven't found it. Essentially, when invoking fetch often times you might want to pass an AbortController signal as a fetch option. I'm currently using `useManualQuery` and ideally I'd like to invalidate a previous request and abort it when the same query is invoked again when in-flight. However, I only see the option to pass fetchOptions to the GraphQLClient on instantiation, and as such this doesn't appear to be supported.

Would there be any way to provide an option to throw out old query results and abort the fetch if a new invocation occurs?

Thanks for your time, have a great day.

### Suggested implementation

I do not know how best to implement this.

kephail:
### What does this PR do?

Create an example of using an abortController. Resolves #1003 

### Checklist


simoneb (changes requested):
- if we're adding a new section to the readme, let's include it in the TOC
- I don't think that using mutation is a sensible example, you would most likely use an abortcontroller to stop a query. stopping a mutation can certainly be done, but it wouldn't give you any guarantees about the outcome of the operation on the server

README.md
```diff
@@ -129,6 +129,7 @@ If you need a client that offers more customization such as advanced cache confi
   - [Testing and mocking](#testing-and-mocking)
   - [Other]
     - [Request interceptors](#request-interceptors)
+    - [abortController](#abortController)
```
simoneb, line 132: please use the right casing. this is a JS class name, and its name is `AbortController`

README.md
```diff
@@ -1258,6 +1259,54 @@ const client = new GraphQLClient({
...
+    await fetchPosts({
+      fetchOptionsOverrides: {
+        signal: controller.signal
+      }
+    })
+  }
+
+  const handleAbort = () => {
+    controller.abort()
+  }
```
simoneb, line 1299: minor: storing values that a React component uses inside module variables is not a very common practice, as you have state lingering around outside of the scope of the component's lifecycle. I understand that this is a simplified example, but since the _correct_ version (which uses Refs) isn't considerably more complex to put in place, I'd go that way to avoid showing misleading code samples

simoneb (changes requested):

README.md
```diff
@@ -1258,6 +1259,54 @@ const client = new GraphQLClient({
...
+### AbortController
+
+if you wish to abort a fetch it is possible to pass an AbortController signal to the `fetchOptionsOverrides` option of the fetch function. This is not `graphql-hooks` specific functionality, rather just an example of how to use it with the library.
+
+```js
+import { GraphQLClient, ClientContext, useManualQuery } from 'graphql-hooks'
+
+const client = new GraphQLClient({
+  url: '/graphql'
+})
```
simoneb, line 1271: remove

README.md
```diff
@@ -1258,6 +1259,54 @@ const client = new GraphQLClient({
...
+    `
+      query {
+        allPosts {
+          id
+          title
+          body
+        }
+      }
+    `
+  )
```
simoneb, line 1285:
```suggestion
  const [fetchData, { loading }] = useManualQuery('...')
```

README.md
```diff
@@ -1258,6 +1259,54 @@ const client = new GraphQLClient({
...
+
+```js
+import { GraphQLClient, ClientContext, useManualQuery } from 'graphql-hooks'
+
+const client = new GraphQLClient({
+  url: '/graphql'
+})
+
+function App() {
+  const abortControllerRef = useRef(null)
```
simoneb, line 1274:
```suggestion
  const abortControllerRef = useRef()
```

README.md
```diff
@@ -1258,6 +1259,54 @@ const client = new GraphQLClient({
...
+          body
+        }
+      }
+    `
+  )
+
+  const handleFetch = () => {
+    abortControllerRef.current = new AbortController()
+    const { signal } = abortControllerRef.current
+    fetchPosts({
```
simoneb, line 1290:
```suggestion
    fetchData({
```

README.md
```diff
@@ -1258,6 +1259,54 @@ const client = new GraphQLClient({
...
+      }
+    })
+  }
+
+  const handleAbort = () => {
+    abortControllerRef.current.abort()
+  }
+
+  return (
+    <ClientContext.Provider value={client}>
```
simoneb, line 1302: remove the wrapping provider

simoneb (approved):

Issue #1007, opened 2023-02-02
 Uncaught TypeError: client.mutationsEmitter.off is not a function


JochenDiekenbrock:
### Package

- [x] `graphql-hooks`
- [ ] `graphql-hooks-ssr`
- [ ] `graphql-hooks-memcache`

### Environment

- `graphql-hooks` version: 6.1.0
- `react` version: preact 10.11.3
- Browser: Chrome/Windows 109, Firefox/Windows 102.7.0esr

### Description

I implemented a web component with preact and graphql-hooks. If the workflow in the web component is completed, I remove it from the DOM. When this is done, an error is displayed in the console:

 Uncaught TypeError: client.mutationsEmitter.off is not a function
    useClientRequest .... rest of stack

I can open a PR, if that's just a missing check in https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks/src/useClientRequest.ts, if client.mutationsEmitter exists, but perhaps I'm just doing something wrong?
The app basically does not much more (graphql wise) than what's described in the Quick start section.

matijagaspar:
@JochenDiekenbrock Could you please provide an example of the code that produces that said issue in preact. Just null guard will not be enough, because the event most likely stays registered and needs clearing to avoid listener leak.

JochenDiekenbrock:
> @JochenDiekenbrock Could you please provide an example of the code that produces that said issue in preact. Just null guard will not be enough, because the event most likely stays registered and needs clearing to avoid listener leak.

Matija, thank you for looking into that! I'm still trying to come up with an example outside of our proprietary code. It was not as straight forward as it seems. Will continue to try...

matijagaspar:
@JochenDiekenbrock I am merging PR: #1011 it has better compatibility across environments and should fix your this issue. Closing this for now, if you still have issues, reopen with examples and as much details as possible: node version, preact version and anything else needed to reproduce it.

PR #1011, opened 2023-02-09
fix: #1007 client.mutationsEmmiter.off is not a function

matijagaspar:
Closes: #1007 

For some reasons the `.off` function was not part of the event emitter in the authors environment, before putting the PR out of WIP need more details from the ticket author, to verify this was the actual problem.


simoneb (approved):

JochenDiekenbrock:
I was busy and could not check the fix earlier, sorry. I've tested now with version 6.2.1 and the error is gone. Great work @matijagaspar , thank you once more!

PR #1013, opened 2023-02-15
chore: ensure generic error type is used in apierror interface

wilkmaia:
### What does this PR do?

Ensures the `TGraphQLError` type is used in the `APIError` generic interface, allowing lib users to change the `graphQLErrors` property type if needed.

This change **should not** break compatibility with previous releases or any change visible to the end user. The `typescript-example` app ran fine with the libs compiled after changes.

I don't believe new documentation or tests are relevant for these changes.

### Related issues

closes https://github.com/nearform/graphql-hooks/issues/973

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/GraphQLClient.ts
+++ b/packages/graphql-hooks/src/GraphQLClient.ts
@@ -154,14 +154,14 @@ class GraphQLClient {
   }
   /* eslint-enable no-console */
 
-  generateResult<ResponseData = any>({
+  generateResult<ResponseData = any, TGraphQLError = object>({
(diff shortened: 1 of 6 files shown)


wilkmaia:
There's a previous PR in place - https://github.com/nearform/graphql-hooks/pull/997 - but it follows a different approach: removing the generic component from the `APIError` interface. This approach, instead, makes use of that generic type and make explicit type calls on functions that relied on the error type being `GraphQLResponseError`.

andysm1th (approved):
LGTM

PR #1015, opened 2023-02-17
fix: import type only to avoid type errors

### LINKED ISSUE #1012: Typescript erroring when graphql-ws and subscriptions-transport-ws aren't installed.
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: 6.1.0
- `react` version: N/A
- Browser: N/A

### Description

Typescript is throwing up errors about `graphql-ws` and `subscriptions-transport-ws` not being present, since they're optional dependencies. The only way I've found to prevent this is to turn on skipLibCheck, or add them as dev-deps, but I'm working in a first party library and I'm concerned that this error would just be shuffled off to the implementing application.

### How to reproduce

This is the `tsconfig` file I'm using.

```json
{
  "include": [
    "src/*.js"
  ],
  "compilerOptions": {
    "allowJs": true,
    "declaration": true,
    "emitDeclarationOnly": true,
    "outDir": "dist",
    "declarationMap": false,
    "sourceMap": false
  }
}
```

The package is just plain JS, i'm generating type definitions from my jsdoc annotations. The error occurs where I import `GraphQLClient`

```
../../node_modules/graphql-hooks/lib/GraphQLClient.d.ts:3:43 - error TS2307: Cannot find module 'graphql-ws' or its corresponding type declarations.

3 import { Client as GraphQLWsClient } from 'graphql-ws';
                                            ~~~~~~~~~~~~

../../node_modules/graphql-hooks/lib/GraphQLClient.d.ts:4:36 - error TS2307: Cannot find module 'subscriptions-transport-ws' or its corresponding type declarations.

4 import { SubscriptionClient } from 'subscriptions-transport-ws';
                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

[If I understand it correctly](https://stackoverflow.com/questions/54392809/how-do-i-handle-optional-peer-dependencies-when-publishing-a-typescript-package#comment123910533_63905376), this is happening because [these two lines](https://github.com/nearform/graphql-hooks/blob/master/packages/graphql-hooks/src/GraphQLClient.ts#L3-L4) aren't explicitly importing just the types. This could also be resolved by adding `// @ts-ignore` above the imports. (I'm fairly green to doing Typescript, so I'm not positive about this).

guilhermelimak:
### What does this PR do?

This will avoid errors when building types in consumer libraries as mentioned in #1012.

### Related issues

 Closes #1012.


simoneb:
Does this close the related issue as well?

simoneb (approved):

simoneb:
Ok, merge away then thanks

PR #1096, opened 2023-07-10
feat: add partial error response mocking

### LINKED ISSUE #1088: How would one mock an error with partial response?
Say we have the following query:

```graphql
query someExample {
  itemA,
  itemB
}
```

I want to write a test case for if `itemB` produces an error, but `itemA` still returns. I'm trying to produce the following response:
```json
{
	"data": {
       "itemA": "foo",
       "itemB": null
    },
	"errors": [
		{
			"message": "Server Error",
			"locations": [
				{
					"line": 5,
					"column": 2
				}
			],
			"path": [
				"itemB"
			]
		}
	]
}
```

Simply returning the `LocalGraphQLError` as it says in the docs causes `data` to be null entirely. Is there a way to attach a data object to the error for return?

matt-clarson:
### What does this PR do?

Adds the ability to mock partial error responses when using `LocalGraphQLClient` by checking the mock response object for keys containing `Error` objects and collecting those errors into the returned error field 

### Related issues

Closes #1088 

### Checklist

### DIFF
--- a/packages/graphql-hooks/src/LocalGraphQLClient.ts
+++ b/packages/graphql-hooks/src/LocalGraphQLClient.ts
@@ -62,7 +62,12 @@ class LocalGraphQLClient extends GraphQLClient {
         if (result instanceof LocalGraphQLError) {
           return { error: result }
         }
-        return { data: result }
+        const {data, errors} = collectErrorsFromObject(result)
+        if (errors.length > 0) {
+          return {data, error: new LocalGraphQLError({graphQLErrors: errors as TGraphQLError[]})}
+        } else {
+          return {data}
+        }
       })
   }
 }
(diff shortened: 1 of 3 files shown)


packages/graphql-hooks/src/LocalGraphQLClient.ts
```diff
@@ -72,4 +77,57 @@ function timeoutPromise(delayInMs) {
...
+      const child = collectErrorsFromObject(entry)
+      data[idx] = child.data
+      errors.push(...child.errors)
+    } else {
+      data[idx] = entry
+    }
+  }
+
+  return {data, errors}
+}
```
drakhart, line 131: I feel like these two functions could be unified somehow, or at least their if-elseif-else logic be extracted to a 3rd function as it is clearly the same for both. The only real differences are the input type, the data init and how the input is traversed.

Everything else looks good to me. 👍

drakhart (changes requested):
As per specific comment.

drakhart (approved):
LGTM!

PR #1105, opened 2023-07-28
fix: useSubscription handles prop changes

### LINKED ISSUE #1094: Subscription won't update on variables change
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: `6.2.3`
  
- `react` version: `18.2.0`
- Browser: Chrome

### Description

When using `useSubscription` and updating its variables (e..g. via `useState`), the subscription won't be updated and will continue to utilize the old values. Related: https://github.com/nearform/graphql-hooks/pull/648.

### How to reproduce

```jsx
  const [stateVariable, setStateVariable] = useState('foo');

  // call setStateVariable somewhere in the component later on

  useSubscription({
    query: subscription,
    variables: {
      variable1: stateVariable,
    },
  });
```

https://github.com/nearform/graphql-hooks/pull/648 should fix this.

matt-clarson:
### What does this PR do?

Adds the `useSubscription` options to the internal `useEffect` dependency array, so that changing the options triggers dropping and recreating a subscription using the new options
One important caveat is adding a custom equality check for the `options.variables` object, to avoid resubscribing when an object equivalent is recreated between renders, e.g.:

```javascript
const MyCoolComponent = () => {
    const [myVar, setMyVar] = useState(123)

    // `variables` here references a different object each rerender,
    // which could cause some fun problems when used with
    // useEffect's dependency array, which uses Object.is to determine equality
    const variables = {myVar}

    useSubscription({query: MY_QUERY, variables}, (...args) => {
        // do stuff
    })
}
```

Other approaches that would work instead of this:
 - Documenting the behaviour outlined in #1094 and including a note in the docs to store the entire variables object in state to avoid `Object.is` issues
 - Something like the [query key concept used by TanStack Query](https://tanstack.com/query/v3/docs/react/guides/query-keys)? 

### Related issues

Closes #1094 

### Checklist


packages/graphql-hooks/src/useSubscription.ts
```diff
@@ -50,8 +62,35 @@ function useSubscription<
...
+ * @example
+ * var o1 = {a: 1, b: 2}
+ * var o2 = {a: 1, b: 2}
+ * isEqual(o1, o2) === true
+ *
+ * var o1 = {a: 2, b: 4}
+ * var o2 = {a: 9, b: 'fish', c: 'why is this here?'}
+ * isEqual(o1, o2) === false
+ */
+function isEqual(o1: object, o2: object): boolean {
```
Puppo, line 81: I'd suggest moving this function in the `packages/graphql-hooks/src/utils.ts` file.

packages/graphql-hooks/test-jsdom/unit/useSubscription.test.tsx
```diff
@@ -283,4 +283,96 @@ describe('useSubscription', () => {
...
+      const result = renderHook(
+        ({ request }) => useSubscription(request, () => {}),
+        { wrapper: Wrapper, initialProps: { request: {query, variables} } }
+      )
+
+      expect(createSubscription).toHaveBeenCalledTimes(1)
+      expect(unsubscribe).toHaveBeenCalledTimes(0)
+
+      result.rerender({
+        request: {query, variables: deepClone(variables)}
```
Puppo, line 349: We can refactor using [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) and remove the deepClone implementation. The CI uses the node lts version, so I don't see any problems on using it.
```ts
        request: {query, variables: structuredClone(variables)}
```

Puppo (changes requested):

Puppo (approved):

simoneb (commented):
I'd recommend that we use this approach instead https://github.com/kentcdodds/use-deep-compare-effect

Puppo:
@simoneb I did the refactor. I noticed that something similar already existed with `useDeepCompareCallback`. So I used your tip about `use-deep-compare-effect` to develop our custom effect hook. Then I refactored the code to share the `useDeepCompareMemoize` in `useDeepCompareEffect` and `useDeepCompareCallback`. I also added tests to cover the new functions.
I'm unsure if adding the refactor in this PR is okay or if creating a new one is better; please, let me know your thoughts. In case splitting the code into two PRs is not a big deal.

simoneb (commented):
is there another place in the code where we're using our own implementation? if so, I would replace that usage as well

simoneb (approved):
LGTM 🚀

PR #1106, opened 2023-08-01
Create scorecard.yml

ShaunBaker:
### What does this PR do?

Adds the [Scorecard Action](https://github.com/ossf/scorecard-action) to execute reviews on the codebase.


simoneb (commented):
I assume that this workflow will be pretty much the same for any repo we want to use it in. If that is the case, I would definitely go with a reusable workflow that we store in a public repo within the org, so that in each repo we just have to set up a simplified workflow which all it does is invoking the shared workflow. How does that sound?

PR #1151, opened 2024-01-16
feat: add support for GQL document input

### LINKED ISSUE #1128: Support TypedDocumentNode
### Description

GraphQL Code Generator has a nifty feature which automatically types (in a Typescript sense) returned code using introspection on the GraphQL server.  However, it requires support for TypedDocumentNode to work.

See https://the-guild.dev/blog/typed-document-node for more details.

[Many other libraries](https://github.com/dotansimha/graphql-typed-document-node#supported-libraries) have already implemented this feature, and it'd be nice if GraphQL-Hooks could add support as well (and get automatic typing).

### Suggested implementation

https://github.com/dotansimha/graphql-typed-document-node#how-can-i-support-this-in-my-library

luke88jones:
### What does this PR do?

- add support for supplying your query in the `DocumentNode` format 
- add support for Typescript response typing via the `TypedDocumentNode` inference 

### Related issues

Resolves #1128 

### Checklist


simoneb:
CI is broken

README.md
```diff
@@ -1364,6 +1364,43 @@ function MyComponent() {
 }
 ```
 
+`graphql-hooks` also supports `TypedDocumentNode`. This allows you to use GraphQL code gen to create `DocumentNode`s for your GQL queries and receive full type support.
```
simoneb, line 1367: let's make sure we also document the simple version without the types, by using `graphql-tag`

examples/typescript/src/App.tsx
```diff
@@ -15,36 +17,36 @@ interface PostData {
 }
 
 const client = new GraphQLClient({
-  url: 'https://create-react-app-server-kqtv5azt3q-ew.a.run.app'
+  url: 'http://localhost:8080/graphql'
```
simoneb, line 20: why this change? I think the idea is that we deploy this as a demo app, IIRC. you will see that there is definitely a graphql server running at that url. I appreciate that this is something that you may want to run locally when you're developing, so let's maybe take the chance to make this dynamic

packages/graphql-hooks/src/useClientRequest.ts
```diff
@@ -87,18 +90,15 @@ function useClientRequest<
...
   UseClientRequestResult<ResponseData, TGraphQLError>,
   ResetFunction
 ] {
-  if (typeof query !== 'string') {
-    throw new Error(
-      'Your query must be a string. If you are using the `gql` template literal from graphql-tag, remove it from your query.'
-    )
-  }
+  const queryString = stringifyDocumentNode(query)
+  const operationName = extractOperationName(query) ?? initialOpts.operationName
```
simoneb, line 101: is this logic correct? if you provide an operation name explicitly, shouldn't it have more priority than the one we infer from the query? unless I'm reading this wrong

packages/graphql-hooks/src/utils.ts
```diff
@@ -1,7 +1,39 @@
...
+
 /**
  * Pipe with support for async functions
  * @param {array} array of functions
  * @returns single function
  */
 export const pipeP = (fns: (() => any)[]) => (arg: any) =>
   fns.reduce((p, f) => p.then(f), Promise.resolve(arg))
+
+export function extractOperationName(document: DocumentNode | string): string | undefined {
```
simoneb, line 16: let's make sure that any new function we create has unit tests please, especially a function like this one that is 1. easy to unit test and 2. has a fairly large set of possible outcomes depending on the input

simoneb (commented):
Good job so far, a few comments inline

examples/typescript/src/App.tsx
luke88jones: There seems to be a mix across the examples. Some use a local GQL server and others connect to this demo. I can look in to making it dynamic if required though

simoneb: we have a demo app running, let's primarily make sure that that one keeps running

packages/graphql-hooks/src/createRefetchMutationsMap.ts
```diff
@@ -0,0 +1,50 @@
+import type { TypedDocumentNode } from '@graphql-typed-document-node/core'
```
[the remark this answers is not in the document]
luke88jones: fixed 👍

README.md
luke88jones: I've added a section on Document node usage

examples/typescript/src/App.tsx
luke88jones: 👍  Reverted the original value

packages/graphql-hooks/src/useClientRequest.ts
luke88jones: Corrected this.

packages/graphql-hooks/src/utils.ts
luke88jones: Tests added.

packages/graphql-hooks/src/useClientRequest.ts
```diff
@@ -1,4 +1,6 @@
 import React from 'react'
+import { TypedDocumentNode } from '@graphql-typed-document-node/core'
```
simoneb, line 2: I have a weird dubt now, after seeing that you added this as a dev dependency, although it's used in production code. unless there's one of the various bundles that we generate which is not actually a bundle but requires the deps to be installed, is there a chance that in fact this package doesn't need any production dependencies whatsoever? if unclear, let's talk :)

luke88jones: 🤔 Yes I think you're right, this should be a production dependency. In the sense that 'production' is someone installing this library to develop with then it'll be required.

simoneb: if somebody is installing this library for developing, will they really need it? on the other hand, if somebody is installing (which is the usual case) this library for _using_ it, I suspect they'll get all of the production deps, which aren't quite necessary I suppose

luke88jones: The type will be required when developing or using the library in order to get the result type information. It is superfluous if the consumer isn't using `DocumentNode`s though

simoneb: ah so it does need to be a production dependency, unless the type information ends up in the bundle, but I suspect it doesn't

simoneb (approved):

Issue #1177, opened 2024-06-01
Graphql dependency doesn't make this a "tiny" bundle anymore


madshall:
### Package

`graphql-hooks`

### Environment

- `graphql-hooks` version: latest

### Description

The latest version has a peer dependency on `graphql` package which is 40kb gzipped. It's 7 times this package's size. So in order to use 
this lightweight package we need to install 40kb additional dependencies. This destroys the whole idea of it.

### Suggested solution (optional)

Make `graphql` an optional dependency or remove the feature that requires it.

carbonrobot:
Consider using [@no-co/graphql.web](https://github.com/0no-co/graphql.web) if this is all client side usage. This is what URQL uses under the covers.

simoneb:
@marceloFerreira90 please take a look

marceloFerreira90:
The reason for the peerDependency of GraphQL is the use of TypedDocumentNode. The approach to fix this will be  to replicate the way it was implemented in graphQL as it only has a dependency on DocumentNode which is available on @no-co/graphql.web

It could be a good issue for that package too.

PR #1178, opened 2024-06-01
feat!: exclude client mutations from refetch condition

madshall:
Client's own options can be changed by a middleware which makes useQuery repeat the call

### What does this PR do?

It excludes client from comparison with a previous value in order to make a decision if a re-fetch is needed. Example: a middleware applies additional headers for specific queries only and restores them back to initial set after the request is complete. This runs `useEffect` hook again and it becomes an infinite loop.

### Related issues

none

### DIFF
--- a/packages/graphql-hooks/src/useQuery.ts
+++ b/packages/graphql-hooks/src/useQuery.ts
@@ -61,7 +61,7 @@ function useQuery<
     setCalledDuringSSR(true)
   }
 
-  const { client, allOptsToStringify } = allOpts
+  const { client: clientFromOpts, allOptsToStringify } = allOpts
   const stringifiedAllOpts = JSON.stringify(allOptsToStringify)
   React.useEffect(() => {
     if (allOpts.skip) {
--- a/packages/graphql-hooks/src/useQuery.ts
+++ b/packages/graphql-hooks/src/useQuery.ts
@@ -61,7 +61,8 @@ function useQuery<
     setCalledDuringSSR(true)
   }
 
-  const stringifiedAllOpts = JSON.stringify(allOpts)
+  const { client, allOptsToStringify } = allOpts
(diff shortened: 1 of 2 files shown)


simoneb:
Thanks for the PR @madshall, and for submitting issue #1177. CI is red due to this change, can you please take a look? Also, it would be great to accompany this change with a test

madshall:
@simoneb I've added the test and fixed the build. Could you please run the pipeline again?

simoneb (commented):
This change makes sense to me, but it's also a breaking change and so I wonder if it makes sense to be included in the library, versus handled in user code. 

For instance, instead of modifying client's headers, can't you create a client with the headers you need?

madshall:
The full use case is this:

Our app talks to multiple graphql servers that's why we don't use a provider, but initialize multiple clients and pass them to useQuery.

We have several middlewares - for dynamic headers (there are some headers which are unique for every request), for logging, error handling etc.

We use useManualQuery everywhere because useQuery gets stuck in a loop when client headers change (after every request).

I understand that technically it is a breaking change, but it really doesn't make sense to trigger a refresh based on allOpts. I'd really only expect it to be triggered when query or variables change, but not anything in the options, however changing that would be really a big deal. Client is the only object in the options that is a class instance and technically it is mutable and shouldn't trigger the hook. 

Would it be better if I put this change behind a feature flag, like an additional option for useQuery? Although it may look ugly having an option like `preventClientMutationsFromTriggeringTheHook`, right? Or maybe a client property, like `immutable=true` could drive this. What do you think?

simoneb:
Ok yes I understand your point. Fair enough, I think we can cut a major just to remain semver compliant, but in practice I expect that this will not affect many users.

simoneb (approved):

PR #1183, opened 2024-06-05
fix: removed dependency on graphql

### LINKED ISSUE #1177: Graphql dependency doesn't make this a "tiny" bundle anymore

marceloFerreira90:
### What does this PR do?

Removed the peer dependency `graphql`
Added dependency on `@0no-co/graphql.web`
Moved typedDocumentNode to inside project (dependency on DocumentNode which is imported from` no-co/graphql `instead of `graphql`

### Related issues

closes #1177

### Checklist

### DIFF
--- a/examples/typescript/src/gql/fragment-masking.ts
+++ b/examples/typescript/src/gql/fragment-masking.ts
@@ -1,86 +1,66 @@
-/* eslint-disable */
-import {
-  ResultOf,
-  DocumentTypeDecoration,
-  TypedDocumentNode
-} from '@graphql-typed-document-node/core'
-import { FragmentDefinitionNode } from 'graphql'
-import { Incremental } from './graphql'
+import { ResultOf, DocumentTypeDecoration, TypedDocumentNode } from '@graphql-typed-document-node/core';
+import { FragmentDefinitionNode } from 'graphql';
+import { Incremental } from './graphql';
 
-export type FragmentType<
-  TDocumentType extends DocumentTypeDecoration<any, any>
-> = TDocumentType extends DocumentTypeDecoration<infer TType, any>
+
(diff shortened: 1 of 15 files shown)


examples/typescript/src/gql/fragment-masking.ts
```diff
@@ -1,66 +1,86 @@
-import { ResultOf, DocumentTypeDecoration, TypedDocumentNode } from '@graphql-typed-document-node/core';
-import { FragmentDefinitionNode } from 'graphql';
-import { Incremental } from './graphql';
+/* eslint-disable */
+import {
+  ResultOf,
+  DocumentTypeDecoration,
+  TypedDocumentNode
+} from '@graphql-typed-document-node/core'
```
simoneb, line 6: wasn't this dependency supposed to be removed?

simoneb (commented):
can you please undo all the formatting changes? they make reviewing this PR quite difficult. just disable it temporarily in your IDE

examples/typescript/src/gql/fragment-masking.ts
marceloFerreira90: this is not within the graphql-hooks package itself this is the example. we still want to allow 'typedDocumentNode'  files being called here but not have a dependency inside the package itself

marceloFerreira90:
closing this pr as my githead is not aligned with master and unable to fix it quickly new pr here: https://github.com/nearform/graphql-hooks/pull/1184

PR #1186, opened 2024-06-10
feat(graphql-hooks): add headers to client response

madshall:
### What does this PR do?

This PR adds `headers` to client response. So along with data, graphQLErrors, httpErrors etc there will be response `headers` as well. Sometimes response headers contain useful info that the client might need to access. Currently the only way to access them is from `responseReducer` because it gets passed the entire response before `generateResult` is called on it. That doesn't scale well because it needs to be set on every single query which leads to repetitive code and lack of unification. One of the use cases for when the response headers are needed is logging middleware:
 - API returns `trace-id` header which can be used in splunk to track down the entire chain of calls to different microservices initiated by the originating client call
 - In order to be able to log this trace id when an error happens we create a logging middleware
 - From there we capture the header value and log the error
Without this change it's impossible to have such a middleware. At least I couldn't find a way.

Let me know if you want me to update the documentation. Didn't do that initially because some teams prefer to do it themselves.

### Related issues

https://github.com/nearform/graphql-hooks/issues/1192

### Checklist


floridemai:
@madshall Thanks for working on this 🙇 
Can you please create an issue where you basically describe the problem to fix (as you did in the description) and link that back here?
Also I've noted a colon at the end of the last property of an object which wasn;t there before. Maybe run an prettier before? It will probably affect other files - up to you if you want to commit those too.

Also, because there was no issue opened. Can we have this as an optional property? Maybe even with a filter on what headers to be returned?

Thank you!

madshall:
@floridemai I've created the issue to reference here, updated the docs and removed the extra colons. Thank you!

floridemai (approved):

madshall:
Hi @simoneb, when can this be published to npm?

simoneb:
Done

PR #1229, opened 2025-01-07
feat: skip option for subscription

### LINKED ISSUE #1225: Add "skip" option to subscription hook
### Description

Add skip option to 'useSubscription' hook.

### Suggested implementation

Add option, default to false. If true, early return before creating observable.

brianjd:
### What does this PR do?

Add skip option for subscriptions, making the creation of subscription conditional.

### Related issues

Closes #1225

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,7 +419,7 @@ To use subscription you can use either [subscriptions-transport-ws](https://gith
 
 `useSubscription(operation, callback)`
 
-- `options`: Object - The GraphQL operation has the following properties:
+- `operation`: Object - The GraphQL operation has the following properties:
   - `query`: String (required) - the GraphQL query
   - `variables`: Object (optional) - Any variables the query might need
   - `operationName`: String (optional) - If your query has multiple operations, you can choose which operation you want to call.
--- a/README.md
+++ b/README.md
@@ -419,11 +419,12 @@ To use subscription you can use either [subscriptions-transport-ws](https://gith
 
 `useSubscription(operation, callback)`
 
-- `operation`: Object - The GraphQL operation has the following properties:
+- `options`: Object - The GraphQL operation has the following properties:
   - `query`: String (required) - the GraphQL query
   - `variables`: Object (optional) - Any variables the query might need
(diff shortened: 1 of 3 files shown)


packages/graphql-hooks/src/useSubscription.ts
```diff
@@ -22,6 +22,10 @@ function useSubscription<
   const contextClient = useContext(ClientContext)
   const client = options.client || contextClient
 
+  if (options.skip) {
+    return
```
simoneb, line 26: although tests are passing, I believe that returning early from a hook violates [React's rules of hooks](https://react.dev/reference/rules/rules-of-hooks), specifically:

> 🔴 Do not call Hooks after a conditional return statement.

brianjd: Ah, you are correct.  I have now moved the skip check inside of the useDeepCompareEffect.

simoneb:
Can you please update the docs to document the option?

https://github.com/nearform/graphql-hooks/?tab=readme-ov-file#useSubscription

brianjd:
@simoneb Done

README.md
```diff
@@ -419,11 +419,12 @@ To use subscription you can use either [subscriptions-transport-ws](https://gith
 
 `useSubscription(operation, callback)`
 
-- `operation`: Object - The GraphQL operation has the following properties:
+- `options`: Object - The GraphQL operation has the following properties:
```
simoneb, line 422: why did you rename this?

brianjd: I renamed because it didn't feel accurate calling it `operation` as it represents more than just the gql operation now.  It was a bit hasty though.  I'll rename it back to `operation`.

simoneb: options is also fine, but let's keep it consistent!

brianjd: Noted.

brianjd: I reverted to `operation` since `options` is being used in other contexts and changing would require more adjustments.

brianjd:
@simoneb Will you be able to craft a release this week with the new option?

simoneb:
@brianjd done

brianjd:
Thanks @simoneb!

PR #1264, opened 2026-08-11
chore(deps): jest 30 + @types/jest 30

lv10:
Follow-up to the consolidated Dependabot PR #1263.

This carries the jest 30 bump from #1254, which could not ride along in the main
consolidation PR because jest 30 is a real migration, not just a version nudge.

Closes #1266

## Included updates

| Package | From | To | Bump | Supersedes |
|---------|------|----|------|------------|
| `jest` | 29.0.1 | 30.2.0 | major | #1254 |
| `@types/jest` | 29.0.3 | 30.0.0 | major | #1254 |
| `babel-jest` | 29.0.1 | 30.2.0 | major | family alignment |
| `jest-environment-jsdom` | 29.0.1 | 30.2.0 | major | family alignment |
| `jest-environment-node` | 29.0.1 | 30.2.0 | major | family alignment |

`ts-jest` is deliberately left at `^29.0.1` — there is no v30, and 29.4.x already declares
a `jest@^30` peer.

## Migration work

1. **Jest family alignment.** Bumping `jest` alone left the satellites at 29, so
   `jest-runtime@30` called into `jest-mock@29` and *every* suite died with
   `TypeError: this._moduleMocker.clearMocksOnScope is not a function`. Fixed by moving
   `babel-jest` / `jest-environment-{jsdom,node}` to 30.

2. **Removed matcher aliases** (jest 30 deletes them) — mechanical rename, 49 call sites
   across 3 test files:
   - `toBeCalledTimes` → `toHaveBeenCalledTimes`
   - `toBeCalledWith` → `toHaveBeenCalledWith`
   - `toBeCalled` → `toHaveBeenCalled`
   - `toThrowError` → `toThrow`

3. **`FormData` cross-realm `Blob` check.** `test-jsdom/unit/GraphQLClient.test.ts`
   imported `File` from `formdata-node` while `GraphQLClient` builds a global **jsdom**
   `FormData`. jsdom 20 accepted the foreign object; jsdom 26 enforces `Blob` branding and
   threw `Failed to execute 'append' on 'FormData': parameter 2 is not of type 'Blob'`.
   Fixed by letting that browser test use the global jsdom `File`; the `formdata-node`
   import is kept where it is intentional (the polyfill scenario in `test-node`).

4. **`window` is non-configurable under jsdom 26.** The test
   `doesn't require fetch to be polyfilled when ssrMode is false and running on the server`
   simulated a server with `delete global.window`, so that `canUseDOM()`
   (`src/canUseDOM.js`, which reads the bare `window` binding) would return false. jsdom 26
   makes that impossible:

   ```
   descriptor: {"configurable":false,"enumerable":true}
   after delete, typeof window: object
   defineProperty -> TypeError: Cannot redefine property: window
   ```

   The delete silently no-opped, `canUseDOM()` stayed true, and with `fetch` deleted the
   constructor threw.

   **Resolved by moving the case to the `graphql-hooks-node` jest project**
   (`testEnvironment: 'node'`), which genuinely has no `window`. The alternative —
   `jest.mock('../../src/canUseDOM')` returning false — would have kept the test in the
   jsdom project but asserted against a stub; a test named "running on the server" is
   better off running somewhere that actually is server-like.

   The moved test also asserts its own premise (`typeof fetch === 'undefined'` after the
   delete). Without that it would pass vacuously: with no DOM and `ssrMode: false`,
   `verifyConfig` skips the fetch check whether or not the delete took effect.

## Build repair (now on `master`)

This branch originally cherry-picked the `typescript` 4.9.5 → 5.9.0 fix from #1263, because
`master` did not build — `typescript@4.9.5` cannot parse the drifted `@types/node`, so
`postinstall` → `build:packages` → `build:types` took `npm install` down with it and CI never
reached the jest changes. Full analysis in #1265.

#1263 has since merged, so that fix is on `master`. This branch is **rebased onto it and the
duplicate commit dropped** — the diff here is now the jest migration and nothing else.

## Verification

Node 24.18.0 (`.nvmrc` → `lts/*`), full `ci.yml` step sequence from a clean tree:

| Step | Result |
|------|--------|
| `npm install` (incl. `postinstall` → `build:packages`) | ✅ |
| `npm run lint` | ✅ |
| `npm run test:coverage -- --silent` | ✅ **225/225**, 23 suites |
| `npm run check-types` | ✅ |
| `npm run build` (all 8 projects) | ✅ |
| `cypress run --spec cypress/e2e/persisted-queries.cy.ts` | ✅ 1/1 (pre-rebase only; covered by `acceptance-tests`) |

Re-run in full on the rebased branch — every step above green again except the local Cypress
spec, which is left to the `acceptance-tests` job.

Test count is unchanged at 225 — the server-side case moved between projects rather than
being dropped. It passes in the `graphql-hooks-node` project and asserts its own premise
(`typeof window === 'undefined'`).


gregoryduckworth (approved):
LGTM 🚀

